PR #2, opened 2023-05-21
Jesse Thomas

jgthomas-12:
I'm surprised I finished this in about 2.5 hours and that it felt rather smooth. I stopped before the extensions so I can go work on my project. 

I messed up calling the params[:id] (in the show method of the recipes(?) controller) at first because I didn't refer to it correctly in the test. Once I remembered how to make that happen correctly, I was able to sort everything else out. 

I'd like to know if I'm using and testing my ingredients model correctly. I'd also like to know if I have any unnecessary testing data in the show spec or how to make any of my tests more robust if they need to be.


app/models/ingredient.rb
```diff
@@ -1,3 +1,10 @@
 class Ingredient < ApplicationRecord
    validates_presence_of :name, :cost
+
+   has_many(:recipe_ingredients)
+   has_many(:recipes, through: :recipe_ingredients)
+
+   def self.calculate_cost
```
cjsim89, line 7: Good use of a class method here. 
I interpreted the user story to ask for `total cost of ingredients **for this recipe**` which would mean an instance method on the `Recipe` model, using the `ingredients` association. That way, the method would only give you the total cost of that recipe's ingredients. 
With your implementation, it would return either ALL ingredients, or the ingredients for a specific collection. Either way is fine, but using a class method can be sometimes a design flaw if used incorrectly. As long as this is tested well, you're good!

config/routes.rb
```diff
@@ -3,4 +3,9 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/ingredients", to: "ingredients#index"
+
+  get "/recipes/:id", to: "recipes#show"
```

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    it "displays a list of all ingredients including their name and cost" do
+      beef = Ingredient.create!(name: "Ground Beef", cost: 2)
+      salt = Ingredient.create!(name: "Salt", cost: 4)
+
+      visit "/ingredients"
+      expect(page). to have_content("Name: #{beef.name}")
+      expect(page). to have_content("Cost: #{beef.cost}")
+
+      expect(page). to have_content("Name: #{salt.name}")
+      expect(page). to have_content("Cost: #{salt.cost}")
```
cjsim89, line 22: This test looks good! I think one way to improve it (though was not required for this IC) would be to use `within` blocks to see that the Beef and Salt ingredients are separated out. 
If we had an id around each ingredient's `<li>` in the view, then we could do something like: 
```
within "#ingredient_#{beef.id}" do
  expect(page).to have_content('"Name: #{beef.name}")
  expect(page).to_not have_content("Name: "#{salt.name}")
end
```

Using something like that would ensure your content is in the right places on the page.

app/views/ingredients/index.html.erb
```diff
@@ -0,0 +1,8 @@
+<h3> Ingredients </h3>
+
+<% @ingredients.each do |ingredient| %>
+  <ul>
```
cjsim89, line 4: This is fine, I would put the `<ul>` element _outside_ the `.each`, that way we only create new `<li>` elements. 
The UL is more about "here's where a list of items will go", and the LI element is for the actual list item itself.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+   it 'displays a recipes name, complexity and genre' do
+      recipe_1 = Recipe.create!(name: "Ground Beef and Salt", complexity: 4, genre: "Impressive")
+      recipe_2 = Recipe.create!(name: "Penut Butter", complexity: 5, genre: "Wicked Impressive")
+
+      visit "/recipes/#{recipe_1.id}"
+      expect(page).to have_content("Recipe Name: #{recipe_1.name}")
+      expect(page).to have_content("Complexity: #{recipe_1.complexity}")
+      expect(page).to have_content("Genre: #{recipe_1.genre}")
+
+      expect(page).not_to have_content("Recipe Name: #{recipe_2.name}")
```
cjsim89, line 21: Great use of `.to_not` here!

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      expect(page).not_to have_content("Complexity: #{recipe_2.complexity}")
+      expect(page).not_to have_content("Genre: #{recipe_2.genre}")
+
+      visit "/recipes/#{recipe_2.id}"
+
+      expect(page).to have_content("Recipe Name: #{recipe_2.name}")
+      expect(page).to have_content("Complexity: #{recipe_2.complexity}")
+      expect(page).to have_content("Genre: #{recipe_2.genre}")
```
cjsim89, line 29: Really well-written and thorough test. 🙌

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+      rec_ing_1 = RecipeIngredient.create!(recipe_id: recipe_1.id, ingredient_id: beef.id)
+      rec_ing_2 = RecipeIngredient.create!(recipe_id: recipe_1.id, ingredient_id: salt.id)
+
+      rec_ing_3 = RecipeIngredient.create!(recipe_id: recipe_2.id, ingredient_id: penuts.id)
+      rec_ing_4 = RecipeIngredient.create!(recipe_id: recipe_2.id, ingredient_id: salt.id)
+
+      visit "/recipes/#{recipe_1.id}"
+      expect(page).to have_content(rec_ing_1.ingredient.name)
+      expect(page).to have_content(rec_ing_2.ingredient.name)
```
cjsim89, line 50: It's fine to do `rec_ing_2.ingredient.name` like you have here, you can also call `salt.name` since you know from this test's data setup which ingredient is which. Not required, but might be easier to type is all.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+      expect(page).not_to have_content(rec_ing_3.ingredient.name)
+      expect(page).not_to have_content(rec_ing_4.ingredient.id)
```
cjsim89, line 53: This specific `expect` line isn't doing much, since we're not displaying the `id` attribute anywhere on the page.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+      visit "/recipes/#{recipe_2.id}"
```
cjsim89, line 55: Good job going to 2 different pages within this test and looking for different content! Very thorough

spec/models/ingredient_spec.rb
```diff
@@ -12,4 +12,17 @@
...
 
+
+   describe "class methods" do
+      describe "#calulate_cost" do
+         it "calculates cost of all ingredients" do
+            beef = Ingredient.create!(name: "Ground Beef", cost: 2)
+            salt = Ingredient.create!(name: "Salt", cost: 4)
+            penuts = Ingredient.create!(name: "Penut", cost: 7)
+
+            expect(Ingredient.calculate_cost).to eq(13)
```
cjsim89, line 23: This test works because you made `calculate_cost` a class method. However you're also using it in the view as part of `@recipe.ingredients`, so it might be good to test that usage in this model test as well. 
Bottom line, it's important that we test model methods in the same way (or ways) that we use them in the actual application.

cjsim89 (commented):
Great work overall, Jesse! Keep up the thorough testing. Remember to come back to finish extensions out for good practice, whenever is convenient for you.

PR #3, opened 2023-05-21
Sarah Garlock

sarahgarlock:
Overall I felt pretty good about this IC. I did spend some time on a specific error on my recipes/show_spec.rb. I'm not sure if I incorrectly set up my relationships but I couldn't connect the recipe.id with the ingredient so I shoveled them in instead. I'd appreciate some feedback about how I can improve this or what error I made. I have also received feedback in the past for not fully testing my order method (Extension 1) so any feedback there would also be appreciated.

### DIFF
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,0 +1,5 @@
+class RecipesController < ApplicationController
+  def show
+    
+  end
+end
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,5 +1,5 @@
 class RecipesController < ApplicationController
   def show
-    
+    @recipe = Recipe.find(params[:id])
   end
 end
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,5 +1,6 @@
 class RecipesController < ApplicationController
   def show
     @recipe = Recipe.find(params[:id])
+    @total_cost = @recipe.ingredients.sum(:cost)
   end
 end
--- a/app/views/ingredients/index.html.erb
+++ b/app/views/ingredients/index.html.erb
@@ -1,6 +1,6 @@
 <h1>Ingredients Index Page</h1>
 <br>
-<% @ingredients.each do |ingredient| %>
+<% @ingredients.order(:name).each do |ingredient| %>
   <h3>Name: <%= ingredient.name %></h3>
   <p>Cost: <%= ingredient.cost %></p>
 <% end %>
--- a/app/views/ingredients/index.html.erb
+++ b/app/views/ingredients/index.html.erb
@@ -1,0 +1,6 @@
+<h1>Ingredients Index Page</h1>
+<br>
+<% @ingredients.each do |ingredient| %>
+  <h3>Name: <%= ingredient.name %></h3>
+  <p>Cost: <%= ingredient.cost %></p>
+<% end %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,4 +5,5 @@ Rails.application.routes.draw do
   # root "articles#index"
 
   get "/ingredients", to: "ingredients#index"
+  get "/recipes/:id", to: "recipes#show"
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,4 +3,6 @@ Rails.application.routes.draw do
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/ingredients", to: "ingredients#index"
 end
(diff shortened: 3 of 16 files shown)


app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,6 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @total_cost = @recipe.ingredients.sum(:cost)
```
cjsim89, line 4: While your ActiveRecord is correct, it doesn't belong in the controller - this should instead be put into a model method (I would say this should go in the Recipe model, and use the `ingredients` association there), and then either we call that method here in the controller, or directly in the view. As it stands, this is considered an MVC infraction.

app/views/ingredients/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1>Ingredients Index Page</h1>
+<br>
+<% @ingredients.order(:name).each do |ingredient| %>
```
cjsim89, line 3: Similar to above, doing `.order` in the view is the job of ActiveRecord, which belongs in the Model. This is an MVC infraction because the View should only be doing presentation logic, while `.order` changes the output of a query. You can confirm this by doing `@ingredients.order(:name).to_sql`, and again without the `.to_sql` to see the query change.

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      expect(page).to have_content(@ingredient_2.name)
+      expect(page).to have_content(@ingredient_2.cost)
+
+      expect(page).to have_content(@ingredient_3.name)
+      expect(page).to have_content(@ingredient_3.cost)
+    end
+
+    it "will display ingredients alphabetically by name" do
+      expect(@ingredient_3.name).to appear_before(@ingredient_1.name)
+      expect(@ingredient_1.name).to appear_before(@ingredient_2.name)
```
cjsim89, line 26: Nice use of `orderly` to make this test work!

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    before(:each) do
+      @ingredient_1 = Ingredient.create!(name: "Ground Beef", cost: 2)
+      @ingredient_2 = Ingredient.create!(name: "Salt", cost: 4)
+      @ingredient_3 = Ingredient.create!(name: "Butter", cost: 5)
+
+      visit "/ingredients"
+    end
+
+    it "will list of all the ingredients including their name and cost" do
+      expect(page).to have_content(@ingredient_1.name)
```
cjsim89, line 14: These tests are ok, but since our view has "Name: " as part of the view that displays this data, we shoudl test for it here. 
```
expect(page).to have_content("Name: "#{ingredient_1.name}")
```

That way, when we do the same thing for Cost, if there were 2 ingredients with the same cost, we're more specific about what "2" is. 
Alternatively (or in addition), you could use a `within `block to be even more specific.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    before(:each) do
+      @recipe_1 = Recipe.create!(name: "Tacos", complexity: 2, genre: "Mexican")
+
+      @ingredient_1 = Ingredient.create!(name: "Ground Beef", cost: 2)
+      @ingredient_2 = Ingredient.create!(name: "Salt", cost: 4)
+      @ingredient_3 = Ingredient.create!(name: "Cheese", cost: 3)
+
+      @recipe_1.ingredients << @ingredient_1
+      @recipe_1.ingredients << @ingredient_2
+      @recipe_1.ingredients << @ingredient_3
```
cjsim89, line 14: This is fine; it's one of a few ways you could create many-to-many instances. Other ways are: 
```
RecipeIngredient.create!(recipe_id: @recipe_1.id, ingredient_id: @ingredient_1
```
or, you could both create and assign a m2m relationship in the same line (I don't like to do this because it's a lot on one line, but it works)
```
@ingredient_1 = @recipe_1.ingredients.create!(name: "Ground Beef", cost: 2)
```

cjsim89: Personally I prefer the first one; `RecipeIngredient.create!(...)` because I can see at a high level what objects belong to which.

config/routes.rb
```diff
@@ -3,4 +3,7 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/ingredients", to: "ingredients#index"
+  get "/recipes/:id", to: "recipes#show"
```
cjsim89, line 8: Good restful routes!

cjsim89 (commented):
Good job overall! Try to focus more on MVC issues - anything to do with data logic should go in a model, and then be tested (and it should be a little alarm in your head if you go through an IC and have *no* model methods); anything that has to do with view logic only should be in the view, and be tested in your feature tests.

PR #4, opened 2023-05-23
Andy Weissman

andyweissman6:
Thought it went ok, had something come up so submitting early.

### DIFF
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,0 +1,6 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @ingredients = @recipe.ingredients
+  end
+end
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -2,5 +2,6 @@ class RecipesController < ApplicationController
   def show
     @recipe = Recipe.find(params[:id])
     @ingredients = @recipe.ingredients
+    @total_cost = @ingredients.sum(:cost)
   end
 end
--- a/app/views/ingredients/index.html.erb
+++ b/app/views/ingredients/index.html.erb
@@ -1,0 +1,6 @@
+<h1>Ingredients Index Page</h1>
+
+<% @ingredients.each do |ingredient| %>
+  <%= ingredient.name %>
+  <%= ingredient.cost %>
+<% end %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,4 +3,7 @@ Rails.application.routes.draw do
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "/ingredients", to: "ingredients#index"
+
+
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,5 +5,8 @@ Rails.application.routes.draw do
   # root "articles#index"
   get "/ingredients", to: "ingredients#index"
 
+  get "/recipes/:id", to: "recipes#show"
+
+
 
 end
--- a/spec/features/ingredients/index_spec.rb
+++ b/spec/features/ingredients/index_spec.rb
@@ -1,10 +1,7 @@
 require "rails_helper"
 
 RSpec.describe "Ingredients Index Page", type: :feature do
-
   let!(:ingredient1) { Ingredient.create!(name: "Peach", cost: "5")}
-
-
 
   describe "ingredients listed" do
     it "should display the list of ingredients including their name and cost" do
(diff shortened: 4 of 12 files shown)


app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,7 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @ingredients = @recipe.ingredients
+    @total_cost = @ingredients.sum(:cost)
```
cjsim89, line 5: While this is correct ActiveRecord, it is querying the database and delivering a calculation, so it should live in a model method, and then that model method should be tested for accuracy. This is considered an MVC infraction.

config/routes.rb
```diff
@@ -3,4 +3,10 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "/ingredients", to: "ingredients#index"
+
+  get "/recipes/:id", to: "recipes#show"
```
cjsim89, line 8: Good restful routes!

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,24 @@
+require "rails_helper"
+
+RSpec.describe "Ingredients Index Page", type: :feature do
+  let!(:ingredient1) { Ingredient.create!(name: "Peach", cost: "5")}
```
cjsim89, line 4: Good job using this `let` block - for feature tests I think it's a good idea to have at least 2-3 objects.

app/views/ingredients/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1>Ingredients Index Page</h1>
+
+<% @ingredients.each do |ingredient| %>
+  <%= ingredient.name %>
+  <%= ingredient.cost %>
```
cjsim89, line 5: I think it might be a good idea to have string labels next to these items; something like 
```
Name: <%= ingredient.name %>
Cost: <%= ingredient.cost %>
```
just so we know what the integer on the page is, and you can have more accurate feature tests for these attributes.

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  describe "ingredients listed" do
+    it "should display the list of ingredients including their name and cost" do
+      visit "/ingredients"
+      
+      expect(page).to have_content("Ingredients Index Page")
+      
+      expect(page).to have_content(ingredient1.name)
+      expect(page).to have_content(ingredient1.cost)
```
cjsim89, line 13: Similar to what I wrote on your view, it would be good to test for `have_content("Cost: #{ingredient1.cost}")`.

cjsim89: Consider a scenario where you have a timestamp on the page somewhere else, and this cost field. 
If the timestamp were 12:00, or 4:12, or even 12:22, and this cost was 12, this test would pass even though it wouldn't be looking directly at the cost field. So, if we test for the full string, it will only try to find that value and ignore the rest.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+# and I see a list of the names of the ingredients for the recipe.
+# `
+require "rails_helper"
+
+RSpec.describe "Recipe Show Page", type: :feature do
+  let!(:ingredient1) { Ingredient.create!(name: "Peach", cost: "3")}
+  let!(:ingredient2) { Ingredient.create!(name: "Cheese", cost: "5")}
+  let!(:recipe1) { Recipe.create!(name: "Burrito", complexity: 3, genre: "Mexican")}
+  let!(:recipe_ingredient1) { recipe1.recipe_ingredients.create!(ingredient: ingredient1)}
+  let!(:recipe_ingredient2) { recipe1.recipe_ingredients.create!(ingredient: ingredient2)}
```
cjsim89, line 15: Much better object setup in this test; and nice job with your m2m objects!

cjsim89: Note: typically we don't need to access the m2m objects by variables, so it can be easier to write something like: 

```
RecipeIngredient.create!(recipe: recipe1, ingrdient: ingredient1)
```
etc. Your way isn't wrong, just more typing.

app/controllers/recipes_controller.rb
cjsim89: A tell-tale sign you may have an MVC infraction in an IC or feature is if you're using activerecord somewhere, and you don't have a model test for it. Tiny spoiler alert - if you finish an IC and _don't_ have any model tests, I would do a quick review!

cjsim89 (commented):
Good work overall, just be careful about writing AR in controllers or views. I would recommend refactoring where I left comments and let us know if you have questions. Also it's a good idea to try those extensions for good practice!

PR #6, opened 2023-05-23
garrettgregor/musical-eureka

garrettgregor:
# Reflection

Overall, I felt pretty good about this and had good momentum going into user story 2.  However, I bogged myself down with alot of test data and could have simplified that quite a bit to make testing easier on myself, because eventually about halfway through the second user story I hit a pretty big roadblock with trying to define my instance method where I kept trying to define it in `RecipeIngredient` model instead of `Recipe` model. This bogged me down quite a bit and made getting through user story 3 pretty stressful. I could definitely use a little more practice with that and how to get the params from forms, as I started to run in to roadblocks with the second extension (albeit with a pretty significant time crunch by the time I got there).

### DIFF
--- a/app/controllers/recipe_ingredients_controller.rb
+++ b/app/controllers/recipe_ingredients_controller.rb
@@ -1,7 +1,6 @@
 class RecipeIngredientsController < ApplicationController
   def update
     redirect = Recipe.find(params[:id])
-    require 'pry'; binding.pry
     recipe = Recipe.create!(name: params[:name], cost: params[:cost])
     RecipeIngredient.create!(recipe_id: redirect.id, ingredient_id: recipe.id)
 
--- a/app/controllers/recipe_ingredients_controller.rb
+++ b/app/controllers/recipe_ingredients_controller.rb
@@ -1,0 +1,11 @@
+class RecipeIngredientsController < ApplicationController
+  def update
+    redirect = Recipe.find(params[:id])
+    require 'pry'; binding.pry
+    recipe = Recipe.create!(name: params[:name], cost: params[:cost])
+    RecipeIngredient.create!(recipe_id: redirect.id, ingredient_id: recipe.id)
+
+    redirect_to "/recipes/#{recipe.id}"
+    recipe_ingredients = RecipeIngredient.find(@recipe.id)
+  end
+end
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,0 +1,6 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @recipe_ingredients = RecipeIngredient.find_ingredients(@recipe.id)
+  end
+end
--- a/app/models/recipe.rb
+++ b/app/models/recipe.rb
@@ -1,3 +1,6 @@
 class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
+
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
 end
--- a/app/models/recipe.rb
+++ b/app/models/recipe.rb
@@ -3,4 +3,8 @@ class Recipe < ApplicationRecord
 
    has_many :recipe_ingredients
    has_many :ingredients, through: :recipe_ingredients
+
+   def total_cost
+      ingredients.sum(:cost)
+   end
 end
--- a/app/models/recipe_ingredient.rb
+++ b/app/models/recipe_ingredient.rb
@@ -1,3 +1,4 @@
 class RecipeIngredient < ApplicationRecord
-
+  belongs_to :recipe
+  belongs_to :ingredient
 end
--- a/app/models/recipe_ingredient.rb
+++ b/app/models/recipe_ingredient.rb
@@ -1,4 +1,8 @@
 class RecipeIngredient < ApplicationRecord
   belongs_to :recipe
   belongs_to :ingredient
+
+  def self.find_ingredients(query_id)
+    where("recipe_id = #{query_id}")
+  end
 end
--- a/app/models/recipe_ingredient.rb
+++ b/app/models/recipe_ingredient.rb
@@ -5,4 +5,5 @@ class RecipeIngredient < ApplicationRecord
   def self.find_ingredients(query_id)
     where("recipe_id = #{query_id}")
   end
+
 end
--- a/app/views/recipes/show.html.erb
+++ b/app/views/recipes/show.html.erb
@@ -1,4 +1,4 @@
-<h1><%= @recipe.name %> Ingredients</h1>
+<h1><%= @recipe.name %></h1>
 <h2>Complexity: <%= @recipe.complexity %></h2>
 <h2>Genre: <%= @recipe.genre %></h2>
 <% @recipe_ingredients.each do |recipe_ingredient| %>
@@ -6,3 +6,4 @@
     <%= recipe_ingredient.ingredient.name %>
   </ul>
 <% end %>
+<h3>Total Cost: $<%= @recipe.total_cost %></h3>
--- a/app/views/recipes/show.html.erb
+++ b/app/views/recipes/show.html.erb
@@ -7,3 +7,14 @@
   </ul>
 <% end %>
 <h3>Total Cost: $<%= @recipe.total_cost %></h3>
+<br>
+<h3>Add Ingredient to <%= @recipe.name %></h3>
+<%= form_with url: "/recipes/#{@recipe.id}", method: :patch, local: true do |f| %>
+  <%= f.label :name %>
+  <%= f.text_field :name%>
+  <br>
+  <%= f.label :cost %>
+  <%= f.number_field :cost %>
+  <br>
+  <%= f.submit "Submit" %>
+<% end %>
--- a/app/views/recipes/show.html.erb
+++ b/app/views/recipes/show.html.erb
@@ -1,0 +1,8 @@
+<h1><%= @recipe.name %> Ingredients</h1>
+<h2>Complexity: <%= @recipe.complexity %></h2>
+<h2>Genre: <%= @recipe.genre %></h2>
+<% @recipe_ingredients.each do |recipe_ingredient| %>
+  <ul>
+    <%= recipe_ingredient.ingredient.name %>
+  </ul>
+<% end %>
--- a/spec/features/ingredients/index_spec.rb
+++ b/spec/features/ingredients/index_spec.rb
@@ -18,5 +18,12 @@ RSpec.describe "/ingredients", type: :feature do
       expect(page).to have_content("#{@ing3.name}")
       expect(page).to have_content("$#{@ing3.cost}")
     end
+
+    it "displays all of the ingredients sorted alphabetically" do
+      visit "/ingredients"
+
+      expect("#{@ing2.name}").to appear_before("#{@ing1.name}")
+      expect("#{@ing1.name}").to appear_before("#{@ing3.name}")
+    end
   end
 end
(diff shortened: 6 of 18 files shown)


app/controllers/recipe_ingredients_controller.rb
```diff
@@ -0,0 +1,10 @@
+class RecipeIngredientsController < ApplicationController
+  def update
+    redirect = Recipe.find(params[:id])
```
cjsim89, line 3: I think I see what you mean here, but I think it's a little dangerous and potentially confusing calling a variable in a controller `redirect`, especially when you're only using it to create a RecipeIngredient.... which is not entirely correct.

app/controllers/recipe_ingredients_controller.rb
```diff
@@ -0,0 +1,10 @@
...
+    recipe = Recipe.create!(name: params[:name], cost: params[:cost])
```
cjsim89, line 4: The 2nd extension asks us to add an existing ingredient to an existing recipe. As such, we don't need to create a new `Recipe` object, just a new relationship between those two existing objects. So, we really only need to `.find` the recipe by its ID, so that we can relate an ingredient and this recipe together.

app/controllers/recipe_ingredients_controller.rb
```diff
@@ -0,0 +1,10 @@
...
+    RecipeIngredient.create!(recipe_id: redirect.id, ingredient_id: recipe.id)
```
cjsim89, line 5: This is the line we're looking for, except you're passing in a `recipe.id` (from line 4) into the `ingredient_id` key; as such this isn't accomplishing what hte user story is looking for, but you are close! 

You'll need a 2nd `.find` method, this time finding an Ingredient by its id. 

With stories like this, it's super important to keep track of any and all Ids you're sending into the controller.

app/controllers/recipe_ingredients_controller.rb
```diff
@@ -0,0 +1,10 @@
...
+
+    redirect_to "/recipes/#{recipe.id}"
+    recipe_ingredients = RecipeIngredient.find(@recipe.id)
```
cjsim89, line 8: Since we're redirecting to a recipe show page, this line will not be used or read.

app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,6 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @recipe_ingredients = RecipeIngredient.find_ingredients(@recipe.id)
```
cjsim89, line 4: I believe this works, but if you're trying to find all the ingredients that belong to this recipe, why not use its association? `@recipe.ingredients` does the same thing for way cheaper.

app/models/recipe.rb
```diff
@@ -1,3 +1,10 @@
 class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
+
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
+
+   def total_cost
+      ingredients.sum(:cost)
```
cjsim89, line 8: Good job making this model method and using this AR! 🙌

app/models/recipe_ingredient.rb
```diff
@@ -1,3 +1,9 @@
 class RecipeIngredient < ApplicationRecord
+  belongs_to :recipe
+  belongs_to :ingredient
+
+  def self.find_ingredients(query_id)
+    where("recipe_id = #{query_id}")
```
cjsim89, line 6: This is a little bit of a code smell, though not super offensively. This reads like you're trying to find all ingredients that belong to a specific recipe, and that's exactly what associations do per instance. But if this is a class method, maybe there are other uses...? I'm not seeing them in this IC though.

cjsim89: By "associations per instance", I mean something like
```
@recipe.ingredients
```

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,20 @@
+<h1><%= @recipe.name %></h1>
+<h2>Complexity: <%= @recipe.complexity %></h2>
+<h2>Genre: <%= @recipe.genre %></h2>
+<% @recipe_ingredients.each do |recipe_ingredient| %>
```
cjsim89, line 4: Like I said elsewhere, this could also be written as 
```
@recipe.ingredients.each do |ingredient|
```
That way you're accessing ingredient objects directly, and don't need to do `recipe_ingreidnet.ingredient.name` below.

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,20 @@
...
+  <ul>
+    <%= recipe_ingredient.ingredient.name %>
+  </ul>
+<% end %>
+<h3>Total Cost: $<%= @recipe.total_cost %></h3>
+<br>
+<h3>Add Ingredient to <%= @recipe.name %></h3>
+<%= form_with url: "/recipes/#{@recipe.id}", method: :patch, local: true do |f| %>
```
cjsim89, line 12: This is close! The 2nd extension asks us to create a new relationship between an existing recipe (from this show page) and an ingredient by its id. As such, the path would not be to a recipe, but rather to either
```
/recipe/:id/ingredients
```
or
```
/recipe_ingredients
```
since those are the relationships or models we'll be affecting the most. 
Also, since we're *creating* a new relationship, we'll use the method `:post`.

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,20 @@
...
+  <%= f.label :name %>
+  <%= f.text_field :name%>
```
cjsim89, line 14: Double-check the user story for extension # 2 - does it ask you to create a new recipe or ingredient? If so, this form setup would be fine. But it asks us to input an ingredient id, and that's all the form should need.

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,29 @@
+require "rails_helper"
+
+RSpec.describe "/ingredients", type: :feature do
+  before(:each) do
+    @ing1 = Ingredient.create!(name: "Lemon", cost: 2)
+    @ing2 = Ingredient.create!(name: "Asparagus", cost: 1)
+    @ing3 = Ingredient.create!(name: "Salmon", cost: 10)
```
cjsim89, line 7: Good object setup for this feature test!

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  describe "ingredients index" do
+    it "displays all of the ingredients" do
+      visit "/ingredients"
+
+      expect(page).to have_content("#{@ing1.name}")
```
cjsim89, line 14: Your view has `"Complexity: ..."` and "Genre" labels, your feature test should too.

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      expect(page).to have_content("$#{@ing2.cost}")
+      expect(page).to have_content("#{@ing3.name}")
+      expect(page).to have_content("$#{@ing3.cost}")
+    end
+
+    it "displays all of the ingredients sorted alphabetically" do
+      visit "/ingredients"
+
+      expect("#{@ing2.name}").to appear_before("#{@ing1.name}")
+      expect("#{@ing1.name}").to appear_before("#{@ing3.name}")
```
cjsim89, line 26: Good job using orderly for this test! Makes it a lot easier. 
Note that you don't need to use the strings and interpolation when using orderly in this way; can just be
```
expect(@ing2.name).to appear_before(@ing1.name)
```
It will assume you mean strings when using capybara.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+    RecipeIngredient.create!(recipe_id: @recipe_2.id, ingredient_id: @ingredient_6.id)
+    RecipeIngredient.create!(recipe_id: @recipe_2.id, ingredient_id: @ingredient_7.id)
+    RecipeIngredient.create!(recipe_id: @recipe_2.id, ingredient_id: @ingredient_8.id)
+
+    # Avo Asparagus
+    RecipeIngredient.create!(recipe_id: @recipe_3.id, ingredient_id: @ingredient_1.id)
+    RecipeIngredient.create!(recipe_id: @recipe_3.id, ingredient_id: @ingredient_2.id)
+    RecipeIngredient.create!(recipe_id: @recipe_3.id, ingredient_id: @ingredient_7.id)
+    RecipeIngredient.create!(recipe_id: @recipe_3.id, ingredient_id: @ingredient_8.id)
+    RecipeIngredient.create!(recipe_id: @recipe_3.id, ingredient_id: @ingredient_9.id)
```
cjsim89, line 41: Woweeeee you were right, that's lots of objects! 

I suggest limiting yourself to 2-3 of each objects maximum for ICs like this. That is usually plenty. However, this is great to see you have the ability to set up this ..magnitude of objects!

spec/models/recipe_ingredients_spec.rb
```diff
@@ -6,4 +6,56 @@
...
+    end
+
+    it "find all ingredients for a recipe" do
+      expected_1 = [@rec_ing_1, @rec_ing_2, @rec_ing_3, @rec_ing_4, @rec_ing_5, @rec_ing_6]
+      expected_2 = [@rec_ing_7, @rec_ing_8, @rec_ing_9, @rec_ing_10, @rec_ing_11]
+      expected_3 = [@rec_ing_12, @rec_ing_13, @rec_ing_14, @rec_ing_15, @rec_ing_16]
+
+      expect(RecipeIngredient.find_ingredients(@recipe_1.id)).to eq(expected_1)
+      expect(RecipeIngredient.find_ingredients(@recipe_2.id)).to eq(expected_2)
+      expect(RecipeIngredient.find_ingredients(@recipe_3.id)).to eq(expected_3)
```
cjsim89, line 58: The danger of adding this model method to your model instead of relying on associations is that you have to do model tests like these, that are already solved for by _using_ those associations. If you find yourself writing tests like these, remember to double-check your associations and let Rails do the work for you!

spec/models/recipe_spec.rb
```diff
@@ -12,4 +12,52 @@
...
+      @rec_ing_13 = RecipeIngredient.create!(recipe_id: @recipe_3.id, ingredient_id: @ingredient_2.id)
+      @rec_ing_14 = RecipeIngredient.create!(recipe_id: @recipe_3.id, ingredient_id: @ingredient_7.id)
+      @rec_ing_15 = RecipeIngredient.create!(recipe_id: @recipe_3.id, ingredient_id: @ingredient_8.id)
+      @rec_ing_16 = RecipeIngredient.create!(recipe_id: @recipe_3.id, ingredient_id: @ingredient_9.id)
+    end
+
+    it "returns the sum of the cost of all ingredients" do
+      expect(@recipe_1.total_cost).to eq(25)
+      expect(@recipe_2.total_cost).to eq(21)
+      expect(@recipe_3.total_cost).to eq(15)
```
cjsim89, line 60: Great job with this model test!

cjsim89 (commented):
Great work overall!

app/controllers/recipes_controller.rb
garrettgregor: 🤯

app/models/recipe_ingredient.rb
garrettgregor: So would this IC not need any model test besides the `.sum`?

spec/features/ingredients/index_spec.rb
garrettgregor: probably went overboard with these... definitely took up some time writing tests

spec/features/ingredients/index_spec.rb
garrettgregor: ✅

spec/models/recipe_spec.rb
garrettgregor: Thank you!

PR #7, opened 2023-05-23
Bradley Milton

brad2412:
I started great by fixing the previous test but then slowed down and was struggling to start the user stories.  I misread the user story and it took me way too long to realize I read wasn't doing the first user story right.  I still feel like I rushed it and do not feel super confident I completed it right.  Because of that I feel like my user story 2 isn't correct either.  It is just frustrating because I feel like I know how to do it but blank out on steps sometimes and I feel like it takes some time to figure out the correct path.  I don't know if that makes sense or not.  I will add that this past week's project has been a tremendous improvement from the 1st project.  I know with more practice and doing the work I will get it just venting a bit.  I will also say you two are awesome and I appreciate how helpful you are.

### DIFF
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,0 +1,6 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @ingredients = @recipe.ingredients
+  end
+end
--- a/app/views/recipes/show.html.erb
+++ b/app/views/recipes/show.html.erb
@@ -1,0 +1,9 @@
+<h1>Recipe Details</h1>
+
+<p><b>Name: </b><%= @recipe.name %></p>
+<p><b>Complexity: </b><%= @recipe.complexity %></p>
+<p><b>Genre: </b><%= @recipe.genre %></p>
+<h2>Ingredients:</h2>
+  <% @ingredients.each do |ingredient| %>
+    <li><%= ingredient.name %></li>
+  <% end %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,4 +3,8 @@ Rails.application.routes.draw do
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/ingredients", to: "ingredients#index"
+
+  get "/recipes/:id", to: "recipes#show"
 end
--- a/spec/features/index_spec.rb
+++ b/spec/features/index_spec.rb
@@ -1,0 +1,34 @@
+require "rails_helper"
+
+RSpec.describe "Ingredients", type: :feature do
+  describe "Ingredients Index" do
+    it "displays a list of all ingredients with their name and cost" do
+      ingredient1 = Ingredient.create(name: "Ground Beef", cost: 2)
+      ingredient2 = Ingredient.create(name: "Salt", cost: 4)
+
+      visit "/ingredients"
+
+      expect(page).to have_content("#{ingredient1.name}: #{ingredient1.cost}")
+      expect(page).to have_content("#{ingredient2.name}: #{ingredient2.cost}")
+    end
+  end
+end
+# RSpec.describe Ingredient, type: :feature do
+#   before(:each) do
+#     @ingredient1 = Ingredient.create!(name: "Ground Beef", cost: 2)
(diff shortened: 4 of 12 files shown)


config/routes.rb
```diff
@@ -3,4 +3,8 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/ingredients", to: "ingredients#index"
+
+  get "/recipes/:id", to: "recipes#show"
```
cjsim89, line 9: Nice restful routes!

app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,6 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @ingredients = @recipe.ingredients
```
cjsim89, line 4: This is fine; you can also access `@recipes.ingredients` from the view itself, which will help keep your controller methods "skinny".

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,9 @@
+<h1>Recipe Details</h1>
+
+<p><b>Name: </b><%= @recipe.name %></p>
+<p><b>Complexity: </b><%= @recipe.complexity %></p>
+<p><b>Genre: </b><%= @recipe.genre %></p>
+<h2>Ingredients:</h2>
+  <% @ingredients.each do |ingredient| %>
+    <li><%= ingredient.name %></li>
```
cjsim89, line 8: good job with this view; if you use `<li>` elements they should be enclosed in a `<ul>` for bullet points, or `<ol>` for numbered lists. Ex: 
```
<ul> 
  <% @ingredients.each do |ingredient| %>
    <li>...</li>
  <% end>
</ul>
```

spec/features/index_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+RSpec.describe "Ingredients", type: :feature do
+  describe "Ingredients Index" do
+    it "displays a list of all ingredients with their name and cost" do
+      ingredient1 = Ingredient.create(name: "Ground Beef", cost: 2)
+      ingredient2 = Ingredient.create(name: "Salt", cost: 4)
+
+      visit "/ingredients"
+
+      expect(page).to have_content("#{ingredient1.name}: #{ingredient1.cost}")
+      expect(page).to have_content("#{ingredient2.name}: #{ingredient2.cost}")
```
cjsim89, line 12: Overall good feature test! I like that you tested exactly what you ended up writing in the view. It may be useful in the future to separate them out by attribute names, e.g.
```
Name: "#{ingredeint1.name}"
Cost: "#{ingredient1.cost}"
```
that way you know which element is which. What if an ingredient is called "$12 spaghetti" and it costs $12, or some weird combo like that?

spec/features/show_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  describe "Recipes Show" do
+    it "displays the recipe's name, complexity, genre, and ingredient names" do
+      recipe = Recipe.create(name: "Spaghetti Bolognese", complexity: "Medium", genre: "Italian")
+      ingredient1 = Ingredient.create(name: "Ground Beef", cost: 2)
+      ingredient2 = Ingredient.create(name: "Salt", cost: 4)
+      recipe.ingredients << [ingredient1, ingredient2]
+
+      visit "/recipes/#{recipe.id}"
+
+      expect(page).to have_content(recipe.name)
```
cjsim89, line 13: Similar to before, it's better in feature tests to test the whole string that appears on the page... So, 
```
have_content("Name: #{recipe.name}")
```

spec/features/show_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+      expect(page).to have_content(recipe.complexity)
+      expect(page).to have_content(recipe.genre)
+      expect(page).to have_content(ingredient1.name)
```
cjsim89, line 16: This area could benefit from a `within` block, to be sure that only these ingredients are part of this recipe. Alternatively, you could have a 3rd ingredient that's not part of the recipe, and do a `to_not` test: 
```
expect(page).to_not have_content(ingredient_3.name)
```

cjsim89 (commented):
Overall good work! I would encourage you to try that 3rd user story, and the extensions for further practice.

PR #8, opened 2023-05-23
Ashley Plasket 

aplasket:
Finished all 3 user stories.
overall, it went better than I thought it would, i had a little trip up over spelling errors (forgetting an 's', spelling html.erb wrong) and on whether total cost should be a recipe instance method or an ingredient class method but figured it out! I'm very surprised at ability to finish the 3 US pretty quickly.


app/models/recipe.rb
```diff
@@ -1,3 +1,10 @@
 class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
+
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
+
+   def total_cost
+      ingredients.sum(:cost)
```
cjsim89, line 8: Great job with both this model method and the AR inside! 🙌

app/views/ingredients/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1>All Ingredients</h1>
+<% @ingredients.each do |ingredient| %>
+  <ul>
+    <li><%= ingredient.name %>: <%= ingredient.cost %></li>
```
cjsim89, line 4: Overall this is ok; I think it imght be better to have each attribute named on the page, like: 
```
Name: {name}
Cost: {cost}
``` 
that way you can tell them apart easily. 

Also, as a side note, the `<ul>` element should surround the entire each block, since the `<li>` elements are list items (bullet points).

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,11 @@
...
+<p>Complexity: <%= @recipe.complexity %></p>
+<p>Genre: <%= @recipe.genre %></p>
+<b>Ingredients: </b>
+<% @recipe.ingredients.each do |ingredient|%>
+  <ul>
+    <li><%= ingredient.name %>
+  </ul>
+<% end %>
+
+<p>Total Cost: <%= @recipe.total_cost %> </p>
```
cjsim89, line 11: Great job with this view!

config/routes.rb
```diff
@@ -3,4 +3,8 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/ingredients", to: "ingredients#index"
+
+  get "/recipes/:id", to: "recipes#show"
```

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    @ingredient2 = Ingredient.create!(name: "Salt", cost: 4)
+  end
+  
+  describe "as a visitor to /ingredients page" do
+    it "should see a list of all ingredients and their attributes" do
+      visit "/ingredients"
+      
+      expect(page).to have_content("All Ingredients")
+      expect(page).to have_content(@ingredient1.name)
+      expect(page).to have_content(@ingredient1.cost)
```
cjsim89, line 15: Tests for index pages benefit a lot from using `within` blocks; otherwise how can we tell from this test where the `@ingredient1.cost` lives, differentiated from the `@ingredient2.cost`?

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    it "should see the specific recipe and it's attributes and ingredients needed" do
+      visit "/recipes/#{@recipe1.id}"
+
+      
+      expect(page).to have_content("Name: #{@recipe1.name}")
+      expect(page).to have_content("Complexity: #{@recipe1.complexity}")
+      expect(page).to have_content("Genre: #{@recipe1.genre}")
+
+      expect(page).to have_content("Ingredients:")
```
cjsim89, line 27: Overall your feature tests are great, you have good object setup and show a step-by-step thought process which is awesome. 

It's important to remember, though, that whenever we write `expect(page).to have_content...` that capybara doesn't go line-by-line on the page too, it will search _the entire page_ for each line. 
So, just because we're expecting the page to have "Ingredients:" above the names of each ingredient, doesn't mean capybara will test that way. 
Instead, this is a perfect use case for a `within` block. If you had a `<div id="ingredients_list">` on your page, you could use a within block like so: 

```
within "#ingredients_list" do
   expect(page).to have_content(@ingredietn1.name)
   ...
end
```
And it would be a far more accurate & precise test of the view for this feature. Give it a try!

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      expect(page).to have_content(@ingredient1.name)
+      expect(page).to have_content(@ingredient3.name)
+      expect(page).to have_content(@ingredient4.name)
+
+      expect(page).to_not have_content(@recipe2.name)
+      expect(page).to_not have_content(@ingredient2.name)
```
cjsim89, line 33: Great to see you using `.to_not` in your tests! Keep this up

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    it "shows the total cost of all ingredients for the recipe" do
+      visit "/recipes/#{@recipe1.id}"
+        # save_and_open_page
+      expect(page).to have_content("Total Cost: 8")
```
cjsim89, line 39: This is good, might be better to use both `recipe` show pages in this test, just to show different values.

spec/models/recipe_spec.rb
```diff
@@ -12,4 +12,22 @@
...
+      @ingredient3 = Ingredient.create!(name: "Pasta", cost: 5)
+      @ingredient4 = Ingredient.create!(name: "Tomato", cost: 1)
+
+      RecipeIngredient.create!(recipe_id: @recipe1.id, ingredient_id: @ingredient1.id)
+      RecipeIngredient.create!(recipe_id: @recipe1.id, ingredient_id: @ingredient3.id)
+      RecipeIngredient.create!(recipe_id: @recipe1.id, ingredient_id: @ingredient4.id)
+    end
+
+    it "#total_cost" do
+      expect(@recipe1.total_cost).to eq(8)
```
cjsim89, line 30: For model tests I think it's good to test against multiple objects - this is fine, but it would be better to have a 2nd recipe with no ingredients, and make sure that returns `0`.

cjsim89 (commented):
Great work overall!

PR #9, opened 2023-05-23
Austin Woods

boomclear:


app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,6 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @recipe_ingredients = RecipeIngredient.where(recipe: @recipe.id)
```
cjsim89, line 4: This controller is off to a good start, but I'm not sure why we need to find all RecipeIngrdients by a specific Recipe's ID.... this reads to me like we're searching for data that we can already get by an association, like
```
@recipe.ingredients
```

In addition, using `.where` in a controller is considered an MVC infraction, since it's altering the query in a dynamic way - we're not sure what data we'll get back, so we would need to test this query to be sure it is correct. If you do need to use this query, it should be moved to a model method, tested, and then called here in the controller. But overall I don't think you need to do this specific query.

app/models/recipe_ingredient.rb
```diff
@@ -1,3 +1,10 @@
 class RecipeIngredient < ApplicationRecord
+  belongs_to :recipe
+  belongs_to :ingredient
 
+  def self.total_cost(array)
+    array.sum do |ingredient|
```
cjsim89, line 6: Using `.sum` on an array like this is using Ruby to do the job of the database via ActiveRecord. 

Instead, we want to use the `.sum(:field)` method in AR. The `:field` in my example is referencing a specific field on a table. 
Specifically, for user story 3, if you're looknig for the total cost of ingredients on a recipe, I would implement this in the `Recipe` model, and use its `ingredients` association with this method, something like: 
```ruby
# Recipe model
def total_cost
   ingredients.sum(:cost)
end
```

This way we're using the association from this Recipe to its ingredients, and letting the database to the sum calculation for us.

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,10 @@
+<p>Recipe: <%= @recipe.name %></p>
+<p>Complexity: <%= @recipe.complexity %></p>
+<p>Genre: <%= @recipe.genre %></p>
+<% @recipe_ingredients.each do |ri| %>
```
cjsim89, line 4: If we rely on the recipe's `ingredients` associaiton, we can rewrite this to be: 
```
@recipe.ingredients.each do |ingredient|
```
and then the objects we're interating through are ingredients already, so we can call `ingredient.name` instead of `ri.ingredient.name`.

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,10 @@
...
+<p>Ingredient: <%= ri.ingredient.name %></p>
+<% end %>
+<p>Total Cost: <%= @recipe_ingredients.sum do |ri| %>
```
cjsim89, line 7: Again this is using Ruby to do the job of ActiveRecord; but this time we're doing Ruby in the view, which since this bit of Ruby is doing data logic, it should be in the Model.

config/routes.rb
```diff
@@ -3,4 +3,8 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/ingredients", to: "ingredients#index"
+  
+  get "/recipes/:id", to: "recipes#show"
```

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      recipe_ingredients2 = RecipeIngredient.create!(recipe: hamburger, ingredient: salt)
+
+      visit "/recipes/#{hamburger.id}"
+
+      expect(page).to have_content("Ingredient: Ground Beef")
+      expect(page).to have_content("Ingredient: Salt")
+      expect(page).to have_content("Recipe: Hamburger")
+      expect(page).to have_content("Complexity: 2")
+      expect(page).to have_content("Genre: American")
+      expect(page).to have_content("Total Cost: 5")
```
cjsim89, line 19: This line for "total cost" is part of a separate user story, as such I would put it in a separate describe block to be sure you're testing each detail of that user story.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,23 @@
+require "rails_helper"
+
+RSpec.describe "Recipe show page" do 
+  describe "shows recipes by id" do
+    it "shows recipes attributes and ingredients" do 
+      beef = Ingredient.create!(name: "Ground Beef", cost: 4)
+      salt = Ingredient.create!(name: "Salt", cost: 1)
+      hamburger = Recipe.create!(name: "Hamburger", complexity: 2, genre: "American")
+      recipe_ingredients1 = RecipeIngredient.create!(recipe: hamburger, ingredient: beef)
+      recipe_ingredients2 = RecipeIngredient.create!(recipe: hamburger, ingredient: salt)
```
cjsim89, line 10: Good object setup for each feature test!

app/models/recipe_ingredient.rb
cjsim89: I also notice you don't have any model tests beyond what was provided in the base repo. This should be an alarm bell in your head for future assessments - if you finish an IC without any additional model methods or tests, double-check your work for MVC issues!

PR #10, opened 2023-05-23
Jesse Sorman

Jesse193:
This includes an ingredients index page and the show page and can return total cost of recipe.

### DIFF
--- a/app/controllers/recipe_ingredients_controller.rb
+++ b/app/controllers/recipe_ingredients_controller.rb
@@ -1,0 +1,6 @@
+class RecipeIngredientsController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:recipe_id])
+    @ingredients = @recipe.ingredients
+  end
+end
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,0 +1,10 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+  end
+
+  private
+    def recipe_params
+      params.permit(:name, :complexity, :genre)
+    end
+end
--- a/app/models/recipe.rb
+++ b/app/models/recipe.rb
@@ -1,3 +1,5 @@
 class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
 end
--- a/app/models/recipe.rb
+++ b/app/models/recipe.rb
@@ -2,4 +2,8 @@ class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
    has_many :recipe_ingredients
    has_many :ingredients, through: :recipe_ingredients
+
+   def total_cost
+      self.ingredients.sum(:cost)
+   end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,4 +5,5 @@ Rails.application.routes.draw do
   # root "articles#index"
   get "/ingredients", to: "ingredients#index"
   get "/recipes/:id", to: "recipes#show"
+  get "/recipes/:id", to: "recipe_ingredients#show"
 end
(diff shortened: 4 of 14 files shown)


spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    ingredient_3 = recipe_1.ingredients.create!(name: "Buns", cost: 3)
+    visit "/recipes/#{recipe_1.id}"
+    expect(page).to have_content("Cost: 17")
+
+    recipe_2 = Recipe.create!(name: "Ravioli", complexity: 3, genre: "Italian")
+    ingredient_4 = recipe_2.ingredients.create!(name: "Flour", cost: 5)
+    ingredient_5 = recipe_2.ingredients.create!(name: "Water", cost: 0)
+    ingredient_6 = recipe_2.ingredients.create!(name: "Cheese", cost: 6)
+    visit "/recipes/#{recipe_2.id}"
+    expect(page).to have_content("Cost: 11")
```
cjsim89, line 37: I really like these feature tests! Your object setup and your using multiple pages to test different values is great thorough work. Keep this up .

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  # (e.g. "Ground Beef: 2" "Salt: 4")
+
+  it "can display all ingredients" do
+    ingredient_1 = Ingredient.create!(name: "Cheese", cost: 4)
+    ingredient_2 = Ingredient.create!(name: "Burger", cost: 10)
+    ingredient_3 = Ingredient.create!(name: "Buns", cost: 3)
+    visit "/ingredients"
+    expect(current_path).to eq("/ingredients")
+    expect(page).to have_content("Name: Cheese")
+    expect(page).to have_content("Cost: 4")
```
cjsim89, line 16: I think this test could beneift from a `within` block - we want to keep the data for Name and Cost of each ingredient in the same spot, otherwise how can we tell from this test that "Cost: 4" belongs to "cheese" and not "buns"?

config/routes.rb
```diff
@@ -3,4 +3,7 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "/ingredients", to: "ingredients#index"
+  get "/recipes/:id", to: "recipes#show"
+  get "/recipes/:id", to: "recipe_ingredients#show"
```
cjsim89, line 8: Overall these routes are restful, but line 8 is never going to be used because Rails will match the verb (`get`) and the path `"/recipes/:id"` from the top of the file down. 

Also, I'm not sure why this route was built, if we're not using the `recipe_ingredients` controller in any stories. I recommend only building a route when you absolutely need it.

app/models/recipe.rb
```diff
@@ -1,3 +1,9 @@
 class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
+
+   def total_cost
+      self.ingredients.sum(:cost)
```
cjsim89, line 7: Great model method and use of AR here!

app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,10 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+  end
+
+  private
+    def recipe_params
+      params.permit(:name, :complexity, :genre)
```
cjsim89, line 8: Good job creating this strong params method. Again, be careful of what you build and when; only build methods or routes (or anything really) as you need it according to the user stories.

app/controllers/recipe_ingredients_controller.rb
```diff
@@ -0,0 +1,6 @@
+class RecipeIngredientsController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:recipe_id])
+    @ingredients = @recipe.ingredients
```
cjsim89, line 4: The 3 user stories you completed don't call for any `#show` method of this controller - though you may want to look at Extension # 2 which may use this controller for a different action. 

As it stands, be careful of spending time on things in these ICs that don't directly have to do with the user stories at hand.

app/models/recipe.rb
cjsim89: However, I don't see a corresponding model test. Be sure your models are always at 100% coverage - before writing a model method, write the test to be sure what AR you implement is correct!

PR #11, opened 2023-05-23
Lauren Golden 

goldenll:
Reflection: I feel like this IC went well for me. I finished all user stories and both extensions with time remaining so I am proud of my work.


app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,10 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @ingredients = Ingredient.all
+    @total_cost = @recipe.total_cost
```
cjsim89, line 5: This is fine, but you can also call `.total_cost` on your `@ingredients` instance inside the view, to keep your controller methods "skinny".

app/models/ingredient.rb
```diff
@@ -1,3 +1,10 @@
 class Ingredient < ApplicationRecord
    validates_presence_of :name, :cost
-end
\ No newline at end of file
+
+   has_many :recipe_ingredients
+   has_many :recipes, through: :recipe_ingredients
+
+   def self.alphabetical
+      @ordered_ingredients = Ingredient.order(:name)
```
cjsim89, line 8: Typically a model method won't assign an instance variable to be returned, especially within a class method like this one, since the value can always be assigned to a different variable where this method is called. 

Having said that, your AR for this method is right on the money!

app/models/recipe.rb
```diff
@@ -1,3 +1,12 @@
...
    validates_presence_of :name, :complexity, :genre
-end
\ No newline at end of file
+
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
+
+   def total_cost
+      @ingredients = Ingredient.all
+      @total_cost = @ingredients.sum(:cost)
```
cjsim89, line 9: Unfortunately, while the AR you wrote `.sum(:cost)` works, it is using this instance variable `@ingredients` which you've assigned to ALL the ingredients. The user story says we want to find the total cost for all ingredients _on this recipe_. So we don't want ALL ingredients, just the ones related to this recipe. 

You're off to a good start with the instance method rather than a class method, but instead of doing `Ingredient.all`, we should rely on the _association_ on this model called `ingredients` on line 5. If we did, our method would look like this: 

```ruby
def total_cost
   ingredients.sum(:cost)
end
```

cjsim89: I'm seeing a trend from your model methods in using instance variables - try to get out of that habit, and instead replace them with only the queries that give you back the data you need.

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,15 @@
+<h1><%= @recipe.name %></h1>
+  <p>Complexity: <%= @recipe.complexity %> </p>
+  <p>Genre: <%= @recipe.genre %> </p>
+  <h3>Ingredients List: <% @ingredients.each do |ingredient| %> </h3>
+      <p><%= ingredient.name %></p>
+    <% end %>
+
+<h3>Total Cost: <%= @total_cost %> </h3>
+
+<%= form_with url: "/recipes/#{@recipe.id}", method: :patch, local: true do |form| %>
```
cjsim89, line 10: Great job working on this 2nd extension! While this form overall is close, the url and methods you use aren't quite correct... 

The user story asks us to create a form that asks for an ingredient_id (which you have done) and then take that ID along with a reicpe and make a new _relationship_ out of those two items. I don't see a completed route or controller action that accomplishes this which is fine; but I recommend trying to finish this part out to see where it could go. 
Also, when you do, consider what I summarized above - are we _updating_ an existing Recipe, or are we _creating_ a new `RecipeIngredient` instance...? If so, what would change about our url and method in this form?

config/routes.rb
```diff
@@ -3,4 +3,9 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "/ingredients", to: "ingredients#index"
+
+  get "/recipes/:id", to: "recipes#show"
+  patch "/recipes/:id", to: "recipes#update"
```
cjsim89, line 9: Overall good restful routes! I don't agree with line 9 for the 2nd extension, but I think you were halfway through this one.

app/models/ingredient.rb
cjsim89: I also don't see any model tests for this method - make sure your TDD process includes writing model methods first and that your coverage for models is always at 100%.

app/models/recipe.rb
cjsim89: Also missing a model test for this method.

spec/features/recipes/recipes_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    expect(page).to_not have_content("Butter")
+    expect(page).to have_field("Add Ingredient")
+
+    ingredient3 = Ingredient.create!(name: "Butter", cost: 4)
+
+    fill_in "add_ingredient", with: "#{ingredient3.id}"
+    click_on "Submit"
+    visit "/recipes/#{recipe1.id}"
+
+    expect(page).to have_content("Butter")
```
cjsim89, line 47: Your feature test for this story is great! See if you can come back and finish this out for good practice.

spec/features/recipes/recipes_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+# User Story 3 - Total Cost
+  it "displays the total cost of all of the ingredients" do 
+    recipe1 = Recipe.create!(name: "Garlic Pasta", complexity: 2, genre: "Italian")
+    ingredient1 = Ingredient.create!(name: "Pasta", cost: 2)
+    ingredient2 = Ingredient.create!(name: "Garlic", cost: 3)
+
+    visit "/recipes/#{recipe1.id}"
+
+    expect(page).to have_content("Total Cost: 5")
```
cjsim89, line 27: While you're testing the value of the total_cost method in this feature test, it doesn't count as a model test.

spec/features/ingredients/ingredients_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  it "displays ingredients in alphabetical order" do 
+    ingredient1 = Ingredient.create!(name: "Pasta", cost: 2)
+    ingredient2 = Ingredient.create!(name: "Garlic", cost: 3)
+
+    visit "/ingredients"
+    
+    expect(page).to have_content(ingredient2.name)
+    expect(page).to have_content(ingredient1.name)
+    ingredient_names = page.all(".ingredient-name").map(&:text)
+    expect(ingredient_names).to eq(ingredient_names.sort)
```
cjsim89, line 27: Great job with this extension and test - you may want to try using the `orderly` gem to make this test a little easier to write. This is great though!

spec/features/ingredients/ingredients_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+RSpec.describe "the ingredients index" do
+# User Story 1 - Ingredients Index
+  it "displays ingredients" do 
+    ingredient1 = Ingredient.create!(name: "Pasta", cost: 2)
+    ingredient2 = Ingredient.create!(name: "Garlic", cost: 3)
+
+    visit "/ingredients"
+
+    expect(page).to have_content(ingredient1.name)
+    expect(page).to have_content(ingredient1.cost)
```
cjsim89, line 12: You may want to consider using string labels for these attributes, so we (and the user) know waht they correspond to. Ex: 
```
have_content("Name: #{ingredient.name")
have_content("Cost: #{ingredient.cost")
```
And to punch it up even further, using a within block is a great way to keep index pages organized - and a great way to tell that this name and cost both belong to `"#ingredient-#{ingredient1.id}"`.

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <%= form.label :add_ingredient, "Add Ingredient" %>
+  <%= form.text_field :add_ingredient %>
```
cjsim89, line 12: Just a suggestion, you may want to call this field what it is, instead of what it _does_; meaning `:ingredient_id` I think is clearer as to what this variable will contain in your controller.

PR #12, opened 2023-05-23
Kim Bergstrom - Hell's Kitchen IC

kbergstrom78:
Kim Bergstrom

I'm nervous about this IC mainly because it seemed really easy.  I completed all the user stories and was close on the first extension.  Hoping I didn't miss something really basic.


README.md
```diff
@@ -1,3 +1,5 @@
+
+<img width="874" alt="Screenshot 2023-05-21 at 9 22 22 AM" src="https://github.com/kbergstrom78/musical-eureka-1233/assets/124642113/be3d1f50-4632-428c-8d1a-d62d909b7525">
```
cjsim89, line 2: Great job adding this screenshot! Know that this is not required for these ICs, but it's a good practice and cool that you included it.

app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,12 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @ingredients = @recipe.ingredients
+    @total_cost = @ingredients.sum(:cost)
```
cjsim89, line 5: While this AR is correct, it's living inside this controller which is an MVC infraction. AR that queries the database for a calculation or a dynamic data set should be in the model, and it should be tested. 

I would call this method something like `total_cost`, and put essentially the same AR you wrote in that method, and either call it here in the controller or directly in the view, either are fine.

app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+  end
+
+  private
+    def recipe_params
```
cjsim89, line 9: Nice job making this strong params method, but I see it's not being called in this controller. Try to only write the code that is necessary for each user story to work.

app/models/ingredient.rb
```diff
@@ -1,3 +1,9 @@
 class Ingredient < ApplicationRecord
    validates_presence_of :name, :cost
+   has_many :recipe_ingredients
+   has_many :recipes, through: :recipe_ingredients
+
+   def self.order_by_name
+      order(name: :asc)
```
cjsim89, line 7: Nice job with this class method and AR!

app/models/recipe.rb
```diff
@@ -1,3 +1,9 @@
 class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
+
+   def total_cost
+   ingredients.sum(:cost)
```
cjsim89, line 7: ....Oh! You did make a `total_cost` method! 👍 I didn't see this at first. As long as this is being called in your view, then you're on the right track. Just make sure to always _remove_ what you refactor and rely on the new code instead.

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,14 @@
...
+  <p>Complexity: <%= @recipe.complexity %></p>
+  <p>Genre: <%= @recipe.genre %></p>
+  <br>
+  <h2>Ingredients:</h2>
+  <ul>
+  <% @ingredients.each do |ingredient| %>
+    <li><%= ingredient.name %></li>
+  </ul> 
+    <% end %>
+  <p> Total Cost of All Ingredients: $ <%= @total_cost %></p>
```
cjsim89, line 13: Since this `@total_cost` variable is coming frmo the controller, you're not actually using your `total_cost` method on the Recipe model. Be careful if this was a refactor that when you move those pieces around, they all line up again at the end.

config/routes.rb
```diff
@@ -3,4 +3,7 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "/ingredients", to: "ingredients#index"
+
+  get "/recipes/:id", to: "recipes#show"
```
cjsim89, line 8: Good restful routes 👍

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    expect(page).to have_content(garlic.cost)
+    expect(page).to have_content(turkey.name)
+    expect(page).to have_content(turkey.cost)
+    expect(page).to have_content(pasta.name)
+    expect(page).to have_content(pasta.cost)
+  end
+
+  it 'lists all ingredients in alpha order' do
+    visit '/ingredients'
+    expect(Ingredient.order_by_name.to_a).to eq([garlic, turkey, pasta])
```
cjsim89, line 25: As a feature test, we should be using capybara to `expect(page).to have_content(...)`. The test you wrote here on line 25 is more of a model test.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  # When I visit '/recipes/:id',
+  # Then I see the recipe's name, complexity and genre,
+  # and I see a list of the names of the ingredients for the recipe.
+
+  let!(:recipe_1) { Recipe.create!(name: "baked ziti", complexity: 2, genre: "italian") }
+  let!(:recipe_2) { Recipe.create!(name: "tamale pie", complexity: 1, genre: "mexican") }
+  
+  let!(:garlic) { recipe_1.ingredients.create!(name: "garlic", cost: 1) }
+  let!(:turkey) { recipe_1.ingredients.create!(name: "ground turkey", cost: 6) }
+  let!(:pasta) { recipe_1.ingredients.create!(name: "penne", cost: 3) }
```
cjsim89, line 15: Good job with your object setup for these tests!

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  it 'lists the recipe info and the list of ingredients)' do
+    visit "/recipes/#{recipe_1.id}"
+    # save_and_open_page
+
+    within("#recipe-#{recipe_1.id}") do
```
cjsim89, line 21: Really good use of a within block here! 🙌

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    expect(page).to have_content(recipe_1.name)
+    expect(page).to have_content(recipe_1.complexity)
+    expect(page).to have_content(recipe_1.genre)
+    expect(page).to have_content("garlic")
+    expect(page).to have_content("ground turkey")
+    expect(page).to have_content("penne")
```
cjsim89, line 27: One way to make this test even better is to include some `.to_not` lines, like: 
```
expect(page).to_not have_content("something that doesn't belong in this within block")
```

spec/models/ingredient_spec.rb
```diff
@@ -12,4 +12,18 @@
...
+      before(:each) do
+         @garlic = Ingredient.create!(name: "garlic", cost: 1)
+         @turkey = Ingredient.create!(name: "ground turkey", cost: 6)
+         @pasta = Ingredient.create!(name: "penne", cost: 3)
+      end
+
+      it 'displays ingredients in alpha order' do
+         ingredients = Ingredient.order_by_name
+
+         expect(ingredients.pluck(:name)).to eq([@garlic.name, @turkey.name, @pasta.name])
```
cjsim89, line 25: This works, you can also do
```
expect(ingredients).to eq([@garlic, @turkey, @pasta])
```

spec/models/recipe_spec.rb
```diff
@@ -12,4 +12,17 @@
...
+  describe '#total_cost' do
+    let!(:recipe_1) { Recipe.create!(name: "baked ziti", complexity: 2, genre: "italian") }
+    let!(:recipe_2) { Recipe.create!(name: "tamale pie", complexity: 1, genre: "mexican") }
+    
+    let!(:garlic) { recipe_1.ingredients.create!(name: "garlic", cost: 1) }
+    let!(:turkey) { recipe_1.ingredients.create!(name: "ground turkey", cost: 6) }
+    let!(:pasta) { recipe_1.ingredients.create!(name: "penne", cost: 3) }
+
+    it 'calculates total cost of a recipes ingredients' do
+      expect(recipe_1.total_cost).to eq(10)
```
cjsim89, line 25: Great job with your model tests. SInce you have a `recipe_2` here, why not go ahead and do 
```
expect(recipe_2.total_cost).to eq(0)
```
? That way you know your method can return zero as well as non-zero results.

PR #13, opened 2023-05-23
Jeff Nelson

jpnelson85:
I was stuck on User Story 2 for awhile with the test setup.  When things did "click", I had 20 minutes and was able to finish user story 2 and 3.  I probably shouldn't have worked on the IC at 7pm after a full day of class and work on our project.  My mind was pretty spent.  Thankfully I went from feeling defeated/discouraged to feeling upbeat/proud of what I accomplished!

### DIFF
--- a/app/controllers/recipe_ingredients_controller.rb
+++ b/app/controllers/recipe_ingredients_controller.rb
@@ -1,0 +1,6 @@
+class RecipeIngredientsController < ApplicationController
+  def index
+    @ingredients = Ingredient.all
+  end
+
+end
--- a/app/controllers/recipe_ingredients_controller.rb
+++ b/app/controllers/recipe_ingredients_controller.rb
@@ -1,6 +1,3 @@
 class RecipeIngredientsController < ApplicationController
-  def index
-    @ingredients = Ingredient.all
-  end
-
+  
 end
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,0 +1,5 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+  end
+end
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,5 +1,6 @@
 class RecipesController < ApplicationController
   def show
     @recipe = Recipe.find(params[:id])
+    @recipe_ingredients = Ingredient.all
   end
 end
--- a/app/views/recipes/show.html.erb
+++ b/app/views/recipes/show.html.erb
@@ -1,0 +1,3 @@
+<h1><%= @recipe.name %></h1>
+  <p> Complexity(1-10): <%= @recipe.complexity %> </p>
+  <p> <%= @recipe.genre %> </p>
--- a/app/views/recipes/show.html.erb
+++ b/app/views/recipes/show.html.erb
@@ -1,3 +1,6 @@
 <h1><%= @recipe.name %></h1>
   <p> Complexity(1-10): <%= @recipe.complexity %> </p>
   <p> <%= @recipe.genre %> </p>
+  <p> <% @recipe_ingredients.each do |ingredient| %> </p>
+    <p> <%= ingredient.name %> </p>
+    <% end %>
--- a/app/views/recipes/show.html.erb
+++ b/app/views/recipes/show.html.erb
@@ -4,3 +4,5 @@
   <p> <% @recipe_ingredients.each do |ingredient| %> </p>
     <p> <%= ingredient.name %> </p>
     <% end %>
+
+    <p> Total Cost: <%= @recipe_ingredients.sum(:cost)%> </p>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,4 +3,7 @@ Rails.application.routes.draw do
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/ingredients", to: "ingredients#index"
+  
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,5 +5,6 @@ Rails.application.routes.draw do
   # root "articles#index"
 
   get "/ingredients", to: "ingredients#index"
-  
+
+  get "recipes/:id", to: "recipes#show"
 end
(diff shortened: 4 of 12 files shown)


app/controllers/recipe_ingredients_controller.rb
```diff
@@ -0,0 +1,3 @@
+class RecipeIngredientsController < ApplicationController
```
cjsim89, line 1: Good job creating this controller, but I would say try to not spend time on anything that's not needed _until you need it_. That way you make sure your focus is on TDD and fixing each error one at a time.

app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,6 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @recipe_ingredients = Ingredient.all
```
cjsim89, line 4: Your instance variable `@recipe_ingredients` is set to `Ingredient.all`, which is actually pulling ALL records from the `ingredients` table, and _not_ the ones that belong to this recipe. For that, you should use an association on Recipe, like: 
```
@recipe.ingredients
```

Depending on how you used this instance variable, it could be a problem.

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1><%= @recipe.name %></h1>
+  <p> Complexity(1-10): <%= @recipe.complexity %> </p>
+  <p> <%= @recipe.genre %> </p>
+  <p> <% @recipe_ingredients.each do |ingredient| %> </p>
```
cjsim89, line 4: Like I said above, since you set this instance variable to be **all ingredients**, and not the ones that belong to this recipe, this is actually going through every single ingredient in the table instead of the ones that belong to this specific recipe, which doesn't meet the requirement of User Story 2.

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,8 @@
...
+    <p> <%= ingredient.name %> </p>
+    <% end %>
+
+    <p> Total Cost: <%= @recipe_ingredients.sum(:cost)%> </p>
```
cjsim89, line 8: Two issues with this line: 
1. Using `@recipe_ingredients` as you implemented in the controller is actually using all ingredients, which is the same problem as I listed above. 
2. Doing .`sum(:cost)` is the right approach for this query, but we're doing it in the view, which is an MVC infraction. A good rule of thumb is to remember to move any AR to a model method and call that method where it needs to be, and then test that model method as well inside a model test file.

config/routes.rb
```diff
@@ -3,4 +3,8 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/ingredients", to: "ingredients#index"
+
+  get "recipes/:id", to: "recipes#show"
```

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    @ham_sandwich = Recipe.create!(name: "Ham Sandwich", complexity: 1, genre: "Meat Sandwich")
+    @bacon_sandwich = Recipe.create!(name: "Bacon Sandwich", complexity: 1, genre: "Meat Sandwich")
+
+    @lt_recipe = RecipeIngredient.create!(recipe_id: @lt.id, ingredient_id: @tomatoe.id)
+    @lt_recipe2 = RecipeIngredient.create!(recipe_id: @lt.id, ingredient_id: @lettuce.id)
+    @lt_recipe3 = RecipeIngredient.create!(recipe_id: @lt.id, ingredient_id: @bread.id)
+    @ham_sandwich_recipe = RecipeIngredient.create!(recipe_id: @ham_sandwich.id, ingredient_id: @ham.id)
+    @ham_sandwich_recipe2 = RecipeIngredient.create!(recipe_id: @ham_sandwich.id, ingredient_id: @bread.id)
+    @bacon_sandwich_recipe = RecipeIngredient.create!(recipe_id: @bacon_sandwich.id, ingredient_id: @bacon.id)
+    @bacon_sandwich_recipe2 = RecipeIngredient.create!(recipe_id: @bacon_sandwich.id, ingredient_id: @bread.id)
```
cjsim89, line 21: Wow, lots of objects created here! That's great, but we really only usually need 2-3 objects per test. If you feel this took you a while to create, you can probably slim down these objects for the next IC.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    expect(page).to have_content(@lt.complexity)
+    expect(page).to have_content(@lt.genre)
+
+    visit "/recipes/#{@ham_sandwich.id}"
+
+    expect(page).to have_content(@ham_sandwich.name)
+    expect(page).to have_content(@ham_sandwich.complexity)
+    expect(page).to have_content(@ham_sandwich.genre)
+
+    visit "/recipes/#{@bacon_sandwich.id}"
```
cjsim89, line 37: Awesome going to different pages to test their attributes; one thing that would make this stronger is to include some `.to_not`s after each visit as well, to be sure content that _doesn't_ belong on each page isn't actually there. Ex: 
```
visit "/recipes/#{@ham_sandwich.id}"

expect(page).to_not have_content(@lt.name)
```

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    visit "/recipes/#{@ham_sandwich.id}"
+
+    expect(page).to have_content("ham")
+    expect(page).to have_content("bread")
+
+    visit "/recipes/#{@bacon_sandwich.id}"
+
+    expect(page).to have_content("bacon")
+    expect(page).to have_content("bread")
```
cjsim89, line 59: The `.to_not` example I showed above would work even better here, in the list of ingredients. 
As it stands, this test would not have shown you the issue I highlighted in the view, where you're actually loading in ALL ingredients; if you had some `.to_not` lines here, you woul dhave probably spotted the problem. 

Try adding
```
expect(page).to_not have("bacon")
```
on line 46 maybe, and see if it passes or fails, to prove my point... It should fail since bacon isn't part of an `lt`.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  it 'shows total cost of all ingredients in the recipe' do
+    visit "/recipes/#{@lt.id}"
+
+    expect(page).to have_content("Total Cost: 30")
```
cjsim89, line 65: This is another one where visiting multiple pages should have given you a different result per page, since your Recipe objects are set up with different sets of ingredients. What you would have seen though, is that each recipe's ingredients total cost would always come back with 30, since you're using `Ingredient.all`. to calculate the sum.

app/views/recipes/show.html.erb
cjsim89: Finally, when you have put this functionality in a model method, it should be tested. Another good thing to remember is that if you finish an IC without any new model methods, you should go back and double-check you're not making any accidental MVC infractions!

PR #14, opened 2023-05-23
Nick Tassinari IC

NickTassinari:


app/controllers/ingredients_controller.rb
```diff
@@ -0,0 +1,6 @@
+class IngredientsController < ApplicationController
+
+  def index 
+    @ingredients = Ingredient.all.order(:name)
```
cjsim89, line 4: The AR for `.order` is correct, but since it's changing the collection in a dynamic way, it should be tested as a model method - remember that the models are responsible for data calculations & querying the Db directly, not the controller, so this would be an MVC infraction.

app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,6 @@
+class RecipesController < ApplicationController
+  def show 
+    @recipe = Recipe.find(params[:id])
+    @total_cost = @recipe.ingredients.sum(:cost)
```
cjsim89, line 4: Similar to your `IngredientsController`, doing `.sum` here is correct AR, but it is doing a calculation on the database. As such, it should be placed in a model method and tested.

config/routes.rb
```diff
@@ -3,4 +3,7 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get '/ingredients', to: 'ingredients#index'
+
+  get '/recipes/:id', to: 'recipes#show'
```
cjsim89, line 8: Good restful routes 👍

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    # When I visit '/ingredients',
+    # Then I see that the list of ingredients is sorted alphabetically by name
+
+    it 'shows the ingredients sorted alphabetically by name' do 
+      ingredient_1 = Ingredient.create!(name: "Xantham gum", cost: 2)
+      ingredient_2 = Ingredient.create!(name: "Salt", cost: 100)
+
+      visit '/ingredients'
+
+      expect(ingredient_2.name).to appear_before(ingredient_1.name)
```
cjsim89, line 35: Nice job using orderly to help write this test!

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    expect(page).to have_content("Crispy Pork Belly")
+    expect(page).to have_content("Recipe Complexity: 3")
+    expect(page).to have_content("Recipe Genre: Awesome")
+  end
+
+    it 'shows the ingredients used in recipe' do 
+      recipe = Recipe.create!(name: "Crispy Pork Belly", complexity: 3, genre: "Awesome")
+      ingredient_1 = recipe.ingredients.create!(name: "Pork Belly", cost: 11)
+      ingredient_2 = recipe.ingredients.create!(name: "Salt", cost: 100)
```
cjsim89, line 25: Good object setup overall. Might be good to have a 2nd recipe with no ingredients, or only 1, just to be able to set up different scenarios in your tests to get better quality coverage.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      visit "/recipes/#{recipe.id}"
+
+      expect(page).to have_content("Pork Belly")
+      expect(page).to have_content("Salt")
```
cjsim89, line 30: I think you should create a 3rd ingredient that's not related to this recipe, so that you can write this test: 

```
expect(page).to_not have_content(ingredient_3.name)
```

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    it 'shows the total cost of the ingredients' do 
+      recipe = Recipe.create!(name: "Crispy Pork Belly", complexity: 3, genre: "Awesome")
+      ingredient_1 = recipe.ingredients.create!(name: "Pork Belly", cost: 11)
+      ingredient_2 = recipe.ingredients.create!(name: "Salt", cost: 100)
+
+      visit "/recipes/#{recipe.id}"
+
+      expect(page).to have_content("Total Cost: 111")
```
cjsim89, line 40: Similar to above, I would like you to get into the habit of thinking just one step further with your tests - it's clear you have a good understanding of how to write tests and how to implement them, but unclear if you understand _why_ we write tests. Adding a line like I did above would demonstrate QA abilities which are important "attention to detail"-type things that employers like to notice on job interviews. 

So for this, maybe a 2nd recipe that we visit, that has either no ingredients or only 1 ingredient, with a different `total_cost` result.

app/controllers/recipes_controller.rb
cjsim89: Also regarding model tests, a little spoiler/hint for you in future ICs.... if you ever finish an IC and realize that you didn't make any new model tests, that should be an alarm that sounds in your mind to go back and double-check for MVC issues!

PR #15, opened 2023-05-23
K.D. Hubbard

kdhubb:
I was able to get through all extensions in this IC. I felt confident in navigating where to place routes and methods in my models and controllers. I think that waiting until Monday probably made this easier for me. I feel like I've learned a lot in the last week and have gotten more comfortable navigating MVC.

### DIFF
--- a/app/controllers/ingredients_controller.rb
+++ b/app/controllers/ingredients_controller.rb
@@ -1,0 +1,5 @@
+class IngredientsController < ApplicationController
+  def index
+    
+  end
+end
--- a/app/controllers/ingredients_controller.rb
+++ b/app/controllers/ingredients_controller.rb
@@ -1,5 +1,5 @@
 class IngredientsController < ApplicationController
   def index
-    
+    @ingredients = Ingredient.all
   end
 end
--- a/app/controllers/ingredients_controller.rb
+++ b/app/controllers/ingredients_controller.rb
@@ -1,5 +1,5 @@
 class IngredientsController < ApplicationController
   def index
-    @ingredients = Ingredient.all
+    @ingredients = Ingredient.order_by_name
   end
 end
--- a/app/controllers/recipe_ingredients_controller.rb
+++ b/app/controllers/recipe_ingredients_controller.rb
@@ -1,0 +1,6 @@
+class RecipeIngredientsController < ApplicationController
+  def create
+
+    redirect_to "/recipes/#{recipe.id}"
+  end
+end
--- a/app/controllers/recipe_ingredients_controller.rb
+++ b/app/controllers/recipe_ingredients_controller.rb
@@ -1,6 +1,8 @@
 class RecipeIngredientsController < ApplicationController
   def create
-
+    recipe = Recipe.find(params[:id])
+    ingredient = Ingredient.find(params[:ingredient_id])
+    RecipeIngredient.create!(recipe: recipe, ingredient: ingredient)
     redirect_to "/recipes/#{recipe.id}"
   end
 end
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,0 +1,5 @@
+class RecipesController < ApplicationController
+  def show 
+    @recipe = Recipe.find(params[:id])
+  end
+end
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,6 +1,6 @@
 class RecipesController < ApplicationController
   def show 
     @recipe = Recipe.find(params[:id])
-    @ingredients = Recipe.ingredients
+    @ingredients = @recipe.ingredients
   end
 end
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,5 +1,6 @@
 class RecipesController < ApplicationController
   def show 
     @recipe = Recipe.find(params[:id])
+    @ingredients = Recipe.ingredients
   end
 end
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -2,5 +2,6 @@ class RecipesController < ApplicationController
   def show 
     @recipe = Recipe.find(params[:id])
     @ingredients = @recipe.ingredients
+    @recipe_cost = @recipe.recipe_cost
   end
 end
--- a/app/models/recipe.rb
+++ b/app/models/recipe.rb
@@ -1,3 +1,5 @@
 class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
 end
--- a/app/models/recipe.rb
+++ b/app/models/recipe.rb
@@ -2,4 +2,8 @@ class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
    has_many :recipe_ingredients
    has_many :ingredients, through: :recipe_ingredients
+
+   def recipe_cost
+      ingredients.sum(:cost)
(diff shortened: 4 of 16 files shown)


app/controllers/ingredients_controller.rb
```diff
@@ -0,0 +1,5 @@
+class IngredientsController < ApplicationController
+  def index
+    @ingredients = Ingredient.order_by_name
```
cjsim89, line 3: Nice job with this class method!

app/controllers/recipe_ingredients_controller.rb
```diff
@@ -0,0 +1,8 @@
+class RecipeIngredientsController < ApplicationController
+  def create
+    recipe = Recipe.find(params[:id])
+    ingredient = Ingredient.find(params[:ingredient_id])
+    RecipeIngredient.create!(recipe: recipe, ingredient: ingredient)
+    redirect_to "/recipes/#{recipe.id}"
```
cjsim89, line 6: Flawless execution. Really great work on this controller action!

app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,7 @@
+class RecipesController < ApplicationController
+  def show 
+    @recipe = Recipe.find(params[:id])
+    @ingredients = @recipe.ingredients
+    @recipe_cost = @recipe.recipe_cost
```
cjsim89, line 5: This is fine, but you can also access the `ingredients` association in the view without any MVC concerns, as well as the `recipe_cost` method. This would also help keep your controllers "skinny".

app/models/recipe.rb
```diff
@@ -1,3 +1,9 @@
 class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
+
+   def recipe_cost
+      ingredients.sum(:cost)
```
cjsim89, line 7: Great AR! 👍

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,19 @@
...
+<% @ingredients.each do |ingredient| %>
+  <ul>
+    <li> <%= ingredient.name%></li>
+  </ul>
+<% end %>
+
+<p><i>Total cost for all ingredients: <%= @recipe_cost %></i></p>
+
+<h3> Add an ingredient to this recipe: </h3>
+<%= form_with url: "/recipes/#{@recipe.id}", method: :post, local: true do |form| %>
```
cjsim89, line 15: I think this is almost perfect - I think your path can be improved in one of two ways: 
either 
```
# 1. 
url: "/recipes/:id/ingredients"

# or 2. 
url: "/recipe_ingredients"
```
Using # 2 would mean you'd have to include the recipe_id as part of the form body as a hidden field. 

As it stands, the route looks like you're creating (`post`) a new Recipe. If we add ingredients to the end of the path, then it reads like "the ingredients for this specific recipe", which is what we're creating.

cjsim89: On the final, this type of error would count as 1 MVC infraction. But overall I like your implementation of this, just one small tweak to make it perfect!

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+RSpec.describe "ingredients index page", type: :feature do 
+  before(:each) do 
+    @ingredient_1 = Ingredient.create!(name: "TVP", cost: 3)
+    @ingredient_2 = Ingredient.create!(name: "Celery", cost: 1)
+    @ingredient_3 = Ingredient.create!(name: "Carrots", cost: 2)
+  end
+
+  it "displays a list of all ingredients in cluding their name and cost" do 
+    visit ingredients_path
```
cjsim89, line 11: Great job with this route helper method! 🙌

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    expect(page).to have_content("All Ingredients")
+    expect(page).to have_content("Name: #{@ingredient_1.name}")
+    expect(page).to have_content("Name: #{@ingredient_2.name}")
+    expect(page).to have_content("Name: #{@ingredient_3.name}")
+    expect(page).to have_content("Cost: #{@ingredient_3.cost}")
+    expect(page).to have_content("Cost: #{@ingredient_2.cost}")
+    expect(page).to have_content("Cost: #{@ingredient_1.cost}")
```
cjsim89, line 19: I think this is a great place to use a `within` block, in index feature tests... Mostly because with this current setup, how can we tell that line 19 belongs to Ingredient 1, or line 17 belongs to ingredient 3?

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  it "displays ingredients listed in alphabetical order" do 
+    visit ingredients_path 
+    
+    expect("Carrots").to appear_before("Celery")
+    expect("Celery").to appear_before("TVP")
+    expect("TVP").to_not appear_before("Carrots")
```
cjsim89, line 27: Great job using orderly!

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    expect(page).to have_content("#{@ingredient_1.name}")
+    expect(page).to have_content("#{@ingredient_2.name}")
+    expect(page).to have_content("#{@ingredient_3.name}")
+    expect(page).to have_content("#{@ingredient_4.name}")
+    expect(page).to have_content("#{@ingredient_5.name}")
+    expect(page).to have_content("#{@ingredient_6.name}")
+    expect(page).to have_content("#{@ingredient_7.name}")
+    expect(page).to have_content("#{@ingredient_8.name}")
+    expect(page).to have_content("#{@ingredient_9.name}")
+    expect(page).to_not have_content("#{@ingredient_9.cost}")
```
cjsim89, line 34: Great job using `.to_not` - I think this is another place where using `within` blocks could help a lot.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  
+  it "displays total cost of all ingredients in the recipe" do 
+    visit "/recipes/#{@recipe_1.id}"
+    
+    expect(page).to have_content("Total cost for all ingredients: #{@recipe_1.recipe_cost}")
+  end
+  
+  it "has a form to add ingredients to the recipe" do 
+    visit "/recipes/#{@recipe_1.id}"
+save_and_open_page
```
cjsim89, line 45: I think we could add an expect statement here, to make sure `ingredient_10` does not exist on the page before we submit the form with it. That would make this test even better.

spec/models/recipe_spec.rb
```diff
@@ -12,4 +24,10 @@
     it {should have_many :recipe_ingredients}
     it {should have_many(:ingredients).through(:recipe_ingredients)}
   end
+
+  describe "instance methods" do 
+    it "#recipe_cost" do 
+      expect(@recipe_1.recipe_cost).to eq(23)
```
cjsim89, line 30: For model tests it's a good idea to try to test against more than 1 object - maybe a `@recipe_2` that has no ingredients, that returns 0...?

cjsim89 (commented):
Really great work overall!

app/controllers/recipes_controller.rb
kdhubb: Just realizing this now! Thanks for this feedback.

app/views/recipes/show.html.erb
kdhubb: Good to know, thank you for this feedback. That makes sense!

PR #16, opened 2023-05-23
Javen Wilson

javenb022:
Completes up to User Story 3
I felt really good during this. It only took me a little less than 2 hours to complete all 3 user stories which I feel means I'm sitting pretty with these concepts


app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,6 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @ingredients = @recipe.ingredients
```
cjsim89, line 4: Good job using this association on Recipe! This instance variable is fine, but you can also reference the association in the view, and that would keep your controller action "skinny".

app/models/recipe.rb
```diff
@@ -1,3 +1,10 @@
 class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
+
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
+
+   def total_cost
+      ingredients.pluck("cost").sum
```
cjsim89, line 8: This is the right idea, where we want to get the cost column on its own and then sum it together, but the way this `sum` method is implemented is actually a Ruby method - whereas in ActiveRecord it would be
```ruby
ingredients.sum(:cost)
```

Another way to think about using AR vs Ruby is, if you were to put `.to_sql` at the end of this line, would it produce SQL or an error? If it's an error, it's probably not using AR to do the calculation.

config/routes.rb
```diff
@@ -3,4 +3,7 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "/ingredients", to: "ingredients#index"
+
+  get "/recipes/:id", to: "recipes#show"
```
cjsim89, line 8: Good restful routes! 👍

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    @beef = Ingredient.create!(name: "Beef", cost: 4)
+    @potatoes = Ingredient.create!(name: "Potatoes", cost: 6)
+    @salt = Ingredient.create!(name: "Salt", cost: 2)
+  end
+  describe "/ingredients" do
+    it "lists all ingredients including their name and cost" do
+      visit "/ingredients"
+
+      expect(page).to have_content("Name: Beef")
+      expect(page).to have_content("Cost: 4")
```
cjsim89, line 14: Try using a `within` block for these index page feature tests; that way we can lump together all the attributes and values for each object within our view.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+RSpec.describe "Recipe Show Page", type: :feature do
+  before(:each) do
+    @filet_mignon = Recipe.create!(name: "Filet Mignon", complexity: "5", genre: "Fancy")
+    @spaghetti = Recipe.create!(name: "Spaghetti", complexity: 1, genre: "Italian")
+
+    @filet_mignon_meat = @filet_mignon.ingredients.create!(name: "Filet Mignon Meat", cost: 50)
+    @butter = @filet_mignon.ingredients.create!(name: "Butter", cost: 2)
+    @noodles = @spaghetti.ingredients.create!(name: "Noodles", cost: 4)
+    @sauce = @spaghetti.ingredients.create!(name: "Sauce", cost: 3)
```
cjsim89, line 11: Overall good object setup for your feature tests!

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      expect(page).to have_content("Filet Mignon Meat")
+      expect(page).to have_content("Butter")
+    end
+
+    it "displays the total cost of all ingredients in a recipe" do
+      visit "/recipes/#{@filet_mignon.id}"
+      expect(page).to have_content("Total Cost: 52")
+
+      visit "/recipes/#{@spaghetti.id}"
+      expect(page).to have_content("Total Cost: 7")
```
cjsim89, line 31: Definitely like seeing that you're using different pages to test different values; that's a good display of attention to detail. Keep this up!

spec/models/recipe_spec.rb
```diff
@@ -12,4 +12,21 @@
...
+    @filet_mignon_meat = @filet_mignon.ingredients.create!(name: "Filet Mignon Meat", cost: 50)
+    @butter = @filet_mignon.ingredients.create!(name: "Butter", cost: 2)
+    @noodles = @spaghetti.ingredients.create!(name: "Noodles", cost: 4)
+    @sauce = @spaghetti.ingredients.create!(name: "Sauce", cost: 3)
+  end
+
+  describe "#total_cost" do
+    it "calculates the total cost of all ingredients in a recipe" do
+      expect(@filet_mignon.total_cost).to eq(52)
+      expect(@spaghetti.total_cost).to eq(7)
```
cjsim89, line 29: Great model test!

PR #17, opened 2023-05-23
Nick Sacco

sicknacco:
I feel I did pretty well this IC. Moved through it fairly easily for once!


app/models/ingredient.rb
```diff
@@ -1,3 +1,10 @@
 class Ingredient < ApplicationRecord
    validates_presence_of :name, :cost
+
+   has_many :recipe_ingredients
+   has_many :recipes, through: :recipe_ingredients
+
+   def self.alpha_ingredients
+      Ingredient.order(:name)
```
epintozzi, line 8:
```suggestion
      order(:name)
```
It is totally fine to explicitly say `Ingredient` here, but since this is a class method defined as `self.method_name`, you don't HAVE to. Just FYI.

app/models/recipe.rb
```diff
@@ -1,3 +1,10 @@
 class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
+
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
+
+   def total_cost
+      self.ingredients.sum(:cost)
```
epintozzi, line 8:
```suggestion
      ingredients.sum(:cost)
```
Instance methods will always be called on `self` implicitly so we don't need to say it explicitly. It's more conventional in ruby to leave it off if we don't need to say it.

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  describe "When I visit '/ingredients'" do
+    it "I see all ingredient names and costs for each" do
+      visit '/ingredients'
+
+      within "#ingredient_#{@beef.id}" do
+        expect(page).to have_content("Name: #{@beef.name}")
+        expect(page).to have_content("Cost: #{@beef.cost}")
+        expect(page).to_not have_content("Name: #{@cheese.name}")
+      end
```
epintozzi, line 16: Nice use of `within` blocks to verify what we should and shouldn't see on various parts of the page 🔥

spec/models/ingredient_spec.rb
```diff
@@ -12,4 +15,11 @@
       it { should have_many(:recipes).through(:recipe_ingredients) }
    end
 
+   describe "class methods" do
+      describe "#alpha_ingredients" do
+        it "can list all ingredients in alphabetical order" do
+          expect(Ingredient.alpha_ingredients).to eq([@tomato, @bread, @bun, @beef, @cheese])
```
epintozzi, line 21: Before I saw your test data, I was a bit concerned about why tomato would be before bread alphabetically 😂

epintozzi (commented):
Great work Nick! You did miss the very last part of US 2: `and I see a list of the names of the ingredients for the recipe.` so watch out for the details and make sure you're completing the full story before moving on. Otherwise this looks great. Your testing is thorough and everything is organized where I would expect it to be following the MVC design pattern.

PR #18, opened 2023-05-23
Julian

JCBeldo:
This went well. I was done with US1-3 with 1.5 hrs to spare and I did EX1 pretty quickly.  I got stuck and felt rushed on EX2.  Didn't fully complete EX2.

### DIFF
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,0 +1,6 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @ingredients = @recipe.ingredients
+  end
+end
--- a/app/models/ingredient.rb
+++ b/app/models/ingredient.rb
@@ -1,3 +1,6 @@
 class Ingredient < ApplicationRecord
    validates_presence_of :name, :cost
+
+   has_many :recipe_ingredients
+   has_many :recipes, through: :recipe_ingredients
 end
--- a/app/models/ingredient.rb
+++ b/app/models/ingredient.rb
@@ -7,4 +7,8 @@ class Ingredient < ApplicationRecord
    def self.total_cost
       self.sum(:cost)
    end
+
+   def self.order_alpha
+      self.order(:name)
+   end
 end
--- a/app/models/ingredient.rb
+++ b/app/models/ingredient.rb
@@ -3,4 +3,8 @@ class Ingredient < ApplicationRecord
 
    has_many :recipe_ingredients
    has_many :recipes, through: :recipe_ingredients
+
+   def self.total_cost
+      self.sum(:cost)
+   end
 end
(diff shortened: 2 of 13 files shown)


app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,6 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @ingredients = @recipe.ingredients
```
epintozzi, line 4: It would be a little more MVC friendly to just call `@recipe.ingredients` in the view rather than defining an extra instance variable in the controller. We want to keep our controllers as thin as possible (without performing data or action based logic in the view).

app/models/ingredient.rb
```diff
@@ -1,3 +1,14 @@
 class Ingredient < ApplicationRecord
    validates_presence_of :name, :cost
+
+   has_many :recipe_ingredients
+   has_many :recipes, through: :recipe_ingredients
+
+   def self.total_cost
+      self.sum(:cost)
+   end
```
epintozzi, line 9: I interpreted US 2 as saying we want the total cost of ingredients in a specific recipe meaning I'd make this an instance method on the `Recipe` model. You're calculating the total cost of ALL ingredients regardless of recipe. You are using active record correctly, however, so that's great! If you moved this to `Recipe`, it would look something like this:
```ruby
class Recipe < Application Record
  ...relationships and validations...

  def total_cost
    ingredients.sum(:cost)
  end
```

app/views/ingredients/index.html.erb
```diff
@@ -0,0 +1,5 @@
+<h1 style="font-size:50px;"><center>Ingredients</h1>
+
+<% @ingredients.each do |ingredient| %>
+  <li><%= ingredient.name %>: <%= number_to_currency(ingredient.cost) %></li>
```
epintozzi, line 4: Oh I like that you're using the currency helper here! Nice touch!

spec/models/ingredient_spec.rb
```diff
@@ -12,4 +21,13 @@
...
    end
 
+   describe "class methods" do   
+      it "::total_cost" do
+         expect(risotto.ingredients.total_cost).to eq(19)
+      end
+
+      it "::order_alpha" do
+         expect(Ingredient.order_alpha).to eq([ing_1, ing_2, ing_3])
+      end
```
epintozzi, line 31: To ensure this is actually alphabetizing, I might rearrange your test objects a bit just to verify you aren't getting a false positive since you are already creating these in alphabetical order. They way you test this in the feature test looks great though, so it is still being covered which is key.

epintozzi (commented):
Nice work Julian! I think you misinterpreted US 3 a little bit but otherwise this looks good. I have a few additional comments below.

PR #19, opened 2023-05-23
Steven Stackhouse 

stackmm:
I was happy with my performance on the first Independent Challenge of mod 2. I was able to get through all three user stories in about 1.5 hours without any real hangups. I finished extension 1 (sort ingredients alphabetically) with issue, but for some reason, my ingredients index page tests wouldn't work. It has something (I think) to do with Orderly? The appears_before method was returning an "undefined method" warning. Since appears_before is the only way I know of for testing sort_alphabetical, I decided to remove all of the extension 1 code from my IC and submit it with just user stories 1-3 included.

### DIFF
--- a/app/models/recipe.rb
+++ b/app/models/recipe.rb
@@ -3,4 +3,8 @@ class Recipe < ApplicationRecord
    has_many :ingredients, through: :recipe_ingredients
    
    validates_presence_of :name, :complexity, :genre
+
+   def total_cost
+      ingredients.pluck(:cost).sum
+   end
 end
--- a/app/models/recipe.rb
+++ b/app/models/recipe.rb
@@ -1,3 +1,6 @@
 class Recipe < ApplicationRecord
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
+   
    validates_presence_of :name, :complexity, :genre
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,4 +3,5 @@ Rails.application.routes.draw do
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "/ingredients", to: "ingredients#index"
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -4,4 +4,5 @@ Rails.application.routes.draw do
   # Defines the root path route ("/")
   # root "articles#index"
   get "/ingredients", to: "ingredients#index"
+  get "/recipes/:id", to: "recipes#show"
 end
--- a/spec/features/recipes/show_spec.rb
+++ b/spec/features/recipes/show_spec.rb
@@ -45,4 +45,14 @@ RSpec.describe "/recipes/:id Show Page" do
     expect(page).to have_content(@ingredient_2.name)
     expect(page).to_not have_content(@ingredient_3.name)
   end
+
+  it "displays the total cost of all ingredients in the recipe" do
+    visit "/recipes/#{@recipe_1.id}"
+
+    expect(page).to have_content("Total Cost: 12")
+
+    visit "/recipes/#{@recipe_2.id}"
+
+    expect(page).to have_content("Total Cost: 6")
+  end
 end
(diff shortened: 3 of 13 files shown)


app/models/recipe.rb
```diff
@@ -1,3 +1,10 @@
 class Recipe < ApplicationRecord
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
+   
    validates_presence_of :name, :complexity, :genre
+
+   def total_cost
+      ingredients.pluck(:cost).sum
```
cjsim89, line 8: I see what you're going for here - you want to separate the `cost` attribute from each ingredient, then add them all together. However, this is a _Ruby_ way of doing this calculation, when the ActiveRecord method `.sum(:attr)` works much better and at scale. So instead of this, it should be: 
```ruby
ingredients.sum(:cost)
```

config/routes.rb
```diff
@@ -3,4 +3,6 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "/ingredients", to: "ingredients#index"
+  get "/recipes/:id", to: "recipes#show"
```
cjsim89, line 7: Good restful routes!

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  it "displays the recipe's name, complexity and genre" do
+    visit "/recipes/#{@recipe_1.id}"
+
+    expect(page).to have_content(@recipe_1.name)
+    expect(page).to have_content("Complexity: #{@recipe_1.complexity}")
+    expect(page).to have_content("Genre: #{@recipe_1.genre}")
+
+    visit "/recipes/#{@recipe_2.id}"
```
cjsim89, line 27: Good job visiting two different show pages and looking for specific content on both. One thing that could improve these tests is the inclusion of `.to_not` expect statements in addition of the `.to` normal ones. You could use this by visiting one page, expecting that content, and then expecting to _not_ see any other content that should belong on a different page. Essentially the "positive" and "negative" cases. A lot of times implementing these kinds of expect statements can open your eyes to the way your program is actually working, and not just the way you assume it to work. Luckily yours is all good here, but keep this in mind!

cjsim89: Example: 
```ruby
visit "/recipes/#{@recipe_1.id}"

expect(page).to have_content(@recipe_1.name)
expect(page).to_not have_content(@recipe_2.name)
# and another line for recipe_2's ingredients that are different from recipe 1. 

```

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    expect(page).to have_content("Ingredients:")
+    expect(page).to have_content(@ingredient_1.name)
+    expect(page).to have_content(@ingredient_2.name)
+    expect(page).to have_content(@ingredient_3.name)
+
+    visit "/recipes/#{@recipe_2.id}"
+
+    expect(page).to have_content(@ingredient_1.name)
+    expect(page).to have_content(@ingredient_2.name)
+    expect(page).to_not have_content(@ingredient_3.name)
```
cjsim89, line 46: Yes, this is what I was talking about above 👍  Keep this up!

spec/models/recipe_spec.rb
```diff
@@ -12,4 +12,25 @@
...
+
+      RecipeIngredient.create!(recipe: recipe_1, ingredient: ingredient_1)
+      RecipeIngredient.create!(recipe: recipe_1, ingredient: ingredient_2)
+      RecipeIngredient.create!(recipe: recipe_1, ingredient: ingredient_3)
+
+      RecipeIngredient.create!(recipe: recipe_2, ingredient: ingredient_1)
+      RecipeIngredient.create!(recipe: recipe_2, ingredient: ingredient_2)
+
+      expect(recipe_1.total_cost).to eq(12)
+      expect(recipe_2.total_cost).to eq(6)
```
cjsim89, line 33: Really great model tests! I think you could include one more recipe with _no_ ingredients, that way we can confirm that the method you wrote will return 0 and not `[]` or `nil`, etc.

PR #20, opened 2023-05-23
Amy Spears

amspears007:
I couldn't get the Ext 2 done, maybe some feedback on how to do that.  TY!


app/controllers/ingredients_controller.rb
```diff
@@ -0,0 +1,6 @@
+class IngredientsController < ApplicationController
+  def index
+    @ingredients = Ingredient.all
+    @sorted_ingredients = Ingredient.sorts_a_z
```
epintozzi, line 4: Ext 1 essentially overrides US 1 in which case it would've been acceptable to get rid of line 3 altogether. Especially because you are no longer using `@ingredients` in your view. That means line 3 is making a call to our database to get ALL of our ingredients and then do nothing with them. That gets pretty expensive for data that we don't need.

app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,10 @@
+class RecipesController <ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @recipe_cost = Recipe.find(params[:id]).total_cost
```
epintozzi, line 4: I would just do `@recipe.total_cost` in your view. That would not be considered and MVC infraction because the method that does that calculation lives in the Model where it should. 

The other thing that isn't ideal about this is you are reaching in to the database twice. Once for your `find` on line 3 and oncee for `find` line 4. You don't need to find the same recipe twice within the same method.

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,25 @@
...
+</section>
+
+<section class="ingredients-<%= @recipe.id %>">
+<h4> Ingredient List </h4>
+  <ol>
+    <% @recipe.ingredients.each do |ingredient| %>
+      <li><%= ingredient.name %></li>
+    <% end %>
+  </ol>
+<p> Total Cost: <%= @recipe_cost %>
```
epintozzi, line 18: This is where I think you could've/should've just used `@recipe.total_cost` instead of creating its own instance variable.

spec/models/recipe_spec.rb
```diff
@@ -12,4 +27,13 @@
     it {should have_many :recipe_ingredients}
     it {should have_many(:ingredients).through(:recipe_ingredients)}
   end
+
+  describe "Total cost of Ingredients" do
+    it "calculates the total cost of the recipes ingredients" do
+
+      expect(pasta_dish.total_cost).to eq(15)
+      expect(chicken_fingers.total_cost).to eq(11)
+      expect(mac_n_cheese.total_cost).to eq(0)
```
epintozzi, line 36: Oh nice! I like that you're testing what happens if no ingredients have been added to a recipe yet.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  describe "visit '/recipes/:id" do
+    it "US2 I see the recipe's name, complexity and genre, and I see a list of the names of the ingredients for the recipe." do
+      visit "/recipes/#{pasta_dish.id}"
+
+      within("h1")do
+        expect(page).to have_content("Recipe Show Page")
+      end
+
+       within("#recipe_attributes-#{pasta_dish.id}") do
```
epintozzi, line 28: Nice use of `within` blocks in your testing!

epintozzi (commented):
This looks good Amy. I have a few comments about cleaning things up a little bit to make the code cleaner and DRYer. Your testing looks really good as well. Nice use of `within` and I can see where you're starting to think about things like edge cases as well.

epintozzi:
I also think you're on the right track for Ext 2. You have a `create` action in recipe, so you'll need to actually store the new ingredient somehow. You might also consider if it makes more sense to create a new `RecipeIngredientsController` to manage that transaction.

PR #21, opened 2023-05-23
Jeffrey Redish

Jredish11:
Just got to extention 2 at 3 hr mark.  Rendering a form for new pet but didnt have time to create a new path  and render the new ingredient

### DIFF
--- a/app/controllers/ingredients_controller.rb
+++ b/app/controllers/ingredients_controller.rb
@@ -1,5 +1,5 @@
 class IngredientsController < ApplicationController
   def index
-    @ingredients = Ingredient.all
+    @ingredients = Ingredient.order(:name)
   end
 end
--- a/app/controllers/ingredients_controller.rb
+++ b/app/controllers/ingredients_controller.rb
@@ -1,5 +1,5 @@
 class IngredientsController < ApplicationController
   def index
-  
+    @ingredients = Ingredient.all
   end
 end
--- a/app/controllers/ingredients_controller.rb
+++ b/app/controllers/ingredients_controller.rb
@@ -1,0 +1,5 @@
+class IngredientsController < ApplicationController
+  def index
+  
+  end
+end
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,0 +1,6 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @ingredients = @recipe.ingredients
+  end
+end
--- a/app/views/recipes/show.html.erb
+++ b/app/views/recipes/show.html.erb
@@ -16,3 +16,11 @@
 </div>
 
 
+<h1>New Ingredients</h1>
+<%= form_with url: "/recipes/#{@recipe.id}", method: :post, local: true do |f| %>
+  <%= f.label :name %>
+  <%= f.text_field :name %>
+
+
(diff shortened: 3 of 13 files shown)


app/controllers/ingredients_controller.rb
```diff
@@ -0,0 +1,5 @@
+class IngredientsController < ApplicationController
+  def index
+    @ingredients = Ingredient.order(:name)
```
epintozzi, line 3: I would move the logic for getting and ordering by name to the `Ingredient` model and make it a class method. This is a bit too much logic for the controller. That said, the ActiveRecord looks good! Nice work!

app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,6 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @ingredients = @recipe.ingredients
```
epintozzi, line 4: I would probably just call `@recipe.ingredients` in the view. That wouldn't be considered too much logic for the view and then has the added benefit of slimming down this controller.

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,26 @@
+<section><h1>Recipes</h1>
+  <h3><%=@recipe.name %></h3>
+  <p>Complexity: <%= @recipe.complexity %></p>
+  <p>Genre: <%= @recipe.genre %></p>
+</section>
+
+<section>
+  <h4>Ingredients</h4>
+  <p><% @ingredients.each do |ingredient| %></p>
```
epintozzi, line 9: This is where I would've just done `@recipe.ingredients.each...` instead of neededing a second instance variable.

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  <p><%= ingredient.name %></p>
+  <% end %>
+</section>
+
+<div>
+  <p>Total Cost:<%= @ingredients.sum { |i| i.cost } %></p>
```
epintozzi, line 15: We don't want to be performing calculations in our view. I'd write a method in your `Recipe` class to do this calculation and then just call the method in the view. The other thing is you're using plain Ruby here and we should try to use ActiveRecord when we can. It's going to be much more efficient for us. Something like this:
```ruby
class Recipe
  ...relationships and validations here...

  def total_cost
   ingredients.sum(:cost)
  end
end
```
and then in the view you'd do:
```ruby
  <p>Total Cost:<%= @recipe.total_cost %></p>
```

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      expect(page).to have_content("Salt")
+      expect(page).to_not have_content("Pizza")
+      expect(page).to_not have_content("Hard")
+      expect(page).to_not have_content("Japanese")
+    end
+  end
+
+  describe "User story 3" do
+    let!(:ingredient_1) { Ingredient.create!(name: "Ground Beef", cost: "4")}
+    let!(:ingredient_2) { Ingredient.create!(name: "Salt", cost: "18")}
```
epintozzi, line 36: Using the fancy expensive salt I see!

epintozzi (commented):
Nice job Jeff. The biggest things to watch out for are keeping your logic in the models and using ActiveRecord when you can. Nice start on Ext 2 as well!

PR #22, opened 2023-05-23
Jimmy Jurado

jcjurado3:
I felt good about this one. I did have to reference a few repos from the past to confirm the set up for the association table. I did not attempt to complete the form extension but would have liked to. I was able to complete the sorted alpha method. I did use .sort at first and had to troubleshoot and change it to .sorted. I guessed Rails didn't like it since it may have thought it was a ruby method.


app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,6 @@
+class RecipesController < ApplicationController
+  def show 
+    @recipe = Recipe.find(params[:id])
+    @ingredient = RecipeIngredient.where(recipe: @recipe.id)
```
epintozzi, line 4: This works, however it would be little more efficient AND would make your controller a little thinner to just do `@recipe.ingredients` in the view. You will get the same results as your code on line 4.

app/models/ingredient.rb
```diff
@@ -1,3 +1,10 @@
 class Ingredient < ApplicationRecord
    validates_presence_of :name, :cost
+
+   has_many :recipe_ingredients
+   has_many :recipes, through: :recipe_ingredients
+
+   def self.sorted
+      order(:name)
+   end
```
epintozzi, line 9: Good use of ActiveRecord! This method should have a unit test in the model_spec as well.

app/models/recipe_ingredient.rb
```diff
@@ -1,3 +1,8 @@
 class RecipeIngredient < ApplicationRecord
+  belongs_to :recipe
+  belongs_to :ingredient
 
+  def self.total_cost
+    joins(:ingredient).sum("ingredients.cost")
```
epintozzi, line 6: I think you make this a touch harder on yourself than necessary. This method is going to return the total cost of ALL ingredients, not just ingredients in a specific recipe. A more effective approach would be to create an instance method in `Recipe` that calculates the cost of just that recipe's ingredients. Good use of AR here, it's just not quite solving the problem we're working on.

app/views/ingredients/index.html.erb
```diff
@@ -0,0 +1,5 @@
+
+<% @ingredients.sorted.each do |ing| %>
```
epintozzi, line 2: Since you got to this extension, I'd probably override the original `@ingredient` in the controller with `Ingredient.sorted`. This works, but you could get rid of the original too once the User Story changed the requirements.

spec/models/recipe_ingredients_spec.rb
```diff
@@ -6,4 +6,17 @@
...
+    before(:each) do
+      @rec1 = Recipe.create!(name: "Tacos", complexity: 5, genre: "Mexican")
+      @ing1 = Ingredient.create!(name: "Ground Beef", cost: 2)
+      @ing2 = Ingredient.create!(name: "Salt", cost: 4)
+      @recing1 = RecipeIngredient.create!(recipe_id: @rec1.id, ingredient_id: @ing1.id)
+      @recing1 = RecipeIngredient.create!(recipe_id: @rec1.id, ingredient_id: @ing2.id)
+    end
+    it "returns total cost of ingredients" do
+      expect(RecipeIngredient.total_cost).to eq(6)
+    end
```
epintozzi, line 20: I think this is a false positive on the method you wrote. If we only want the total cost of all ingredients for a specific recipe, this will pass because we only have one recipe. If we create another recipe with more ingredients, I think this would fail and would've flagged that the query wasn't quite right.

epintozzi (commented):
Nice job Jimmy! I left some comment regarding general clean up but overall you seem to be on the right track.

app/controllers/recipes_controller.rb
jcjurado3: I get this now, after looking at it later this week. Thank you!

app/models/recipe_ingredient.rb
jcjurado3: Ahhhh I see. Thank you for the feedback. I'll continue to work on when to use instance method and which model would be the appropriate use for that specific user story.

PR #23, opened 2023-05-23
Michael Callahan

calforcal:


app/models/ingredient.rb
```diff
@@ -1,3 +1,14 @@
 class Ingredient < ApplicationRecord
-   validates_presence_of :name, :cost
+  has_many :recipe_ingredients
+  has_many :recipes, through: :recipe_ingredients
+  validates_presence_of :name, :cost
+
+  def self.total_cost
+    @ingredients = Ingredient.all
+    cost = @ingredients.sum { |ingredient| ingredient.cost }
```
cjsim89, line 8: I can understand your logic here, but this is a Ruby method and we want to rely on AR methods instead. The AR equivalent would be `.sum(:cost)`, where .sum is called on a collection of objects, and we pass in a symbol for the name of the attribute we want to sum. 

In addition, we can call this on `Ingredient.all`, or just `Ingredient` if we want to do it on the entire table. 

However, the user story calls for the total cost of the recipe by addnig up all of _that recipe's ingredients_, so this could work better and easier if we implemented this functionality in a Recipe instance method, like this: 
```ruby
def total_cost
   ingredients.sum(:cost) #where ingredients is the association on Recipe
end
```

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,10 @@
+<h1><%= "#{@recipe.name}" %></h1>
+<p><%= "Complexity: #{@recipe.complexity}" %></p>
+<p><%= "Genre: #{@recipe.genre}" %></p>
+<ul><p><%= "Ingredients:" %></p>
+  <% @recipe.ingredients.each do |ingredient| %>
+    <li><%= ingredient.name %></li>
+  <% end %>
+</ul>
+
+<h4><%= "Total Cost: #{@recipe.ingredients.total_cost}" %></h4>
```
cjsim89, line 10: Since you're calling `total_cost` on this collection, it should end up showing the right amounts on this page, but be careful with this kind of design because it could also be called on "all" ingredients which would give you a different amount. As long as this is tested thoroughly, you should be fine.

config/routes.rb
```diff
@@ -3,4 +3,8 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/ingredients", to: "ingredients#index"
+
+  get "/recipes/:recipe_id", to: "recipes#show"
```

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    visit "/ingredients"
+
+    expect(page).to have_content("#{ingredient_1.name}: #{ingredient_1.cost}")
+    expect(page).to have_content("#{ingredient_2.name}: #{ingredient_2.cost}")
+  end
+
+  it "displays the ingredients in alphabetical order" do
+    visit "/ingredients"
+
+    expect(ingredient_2.name).to appear_before(ingredient_1.name)
```
cjsim89, line 17: Good use of orderly for this test!

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  let!(:ingredient_1) { recipe.ingredients.create!(name: "Garlic Powder", cost: 1)}
+  let!(:ingredient_2) { recipe.ingredients.create!(name: "Beans", cost: 3)}
+
+  it "displays a specific recipes name, complexity, genre and list of ingredients" do
+    visit "/recipes/#{recipe.id}"
+
+    expect(page).to have_content("#{recipe.name}")
+    expect(page).to have_content("Complexity: #{recipe.complexity}")
+    expect(page).to have_content("Genre: #{recipe.genre}")
+    expect(page).to have_content("Ingredients:")
```
cjsim89, line 14: This section would be a perfect use case for a `within` block - in fact anywhere you have a `.each` in your views, you could benefit from a within block because all that data to be displayed will be dynamic.

spec/models/ingredient_spec.rb
```diff
@@ -12,4 +12,25 @@
       it { should have_many(:recipes).through(:recipe_ingredients) }
    end
 
+    describe "total_cost" do
+      it "can total the cost of ingredients" do
+        ingredient_1 = Ingredient.create!(name: "Garlic Powder", cost: 1)
+        ingredient_2 = Ingredient.create!(name: "Beans", cost: 3)
+        
+        expect(Ingredient.total_cost).to eq(4)
```
cjsim89, line 20: Since this is a class method and you're using it on an instance's association in your feature test, that you also model that behavior in this model test. I think it would be easier to implement as a Recipe's instance method instead.

spec/models/ingredient_spec.rb
```diff
@@ -12,4 +12,25 @@
...
+    describe "sort_alphabetical" do
+      it "sorts ingredients alphabetically" do
+        ingredient_1 = Ingredient.create!(name: "Garlic Powder", cost: 1)
+        ingredient_2 = Ingredient.create!(name: "Beans", cost: 3)
+
+        original = [ingredient_1, ingredient_2]
+        expected = [ingredient_2, ingredient_1]
+
+        expect(Ingredient.all).to eq(original)
+        expect(Ingredient.sort_alphabetical).to eq(expected)
```
cjsim89, line 33: This is a good way to set up this test!

PR #24, opened 2023-05-23
Kailey Kaes

kaileykaes:
This went better than expected. Encouraging. Got hung up on plurals there for a hot second. Felt a whole lotta NOPE when I read through the second extension, so I wrapped up before time instead.


app/models/ingredient.rb
```diff
@@ -1,3 +1,13 @@
 class Ingredient < ApplicationRecord
-   validates_presence_of :name, :cost
+  validates_presence_of :name, :cost
+  has_many :recipe_ingredients
+  has_many :recipes, through: :recipe_ingredients
+
+  def self.sort_alphabetically(sort)
```
cjsim89, line 6: I see what you're going for here, and it makes sense and is well-implemented. However know for the other ICs that you don't have to implement this sort of pattern, if it asks you to do a specific ordering you can remove the other kind of ordering on the same object. But this looks good!

app/models/recipe.rb
```diff
@@ -1,3 +1,9 @@
 class Recipe < ApplicationRecord
-   validates_presence_of :name, :complexity, :genre
+  validates_presence_of :name, :complexity, :genre
+  has_many :recipe_ingredients
+  has_many :ingredients, through: :recipe_ingredients
+
+  def total_cost
+    ingredients.sum(:cost)
```
cjsim89, line 7: Good AR here!

app/views/ingredients/index.html.erb
```diff
@@ -0,0 +1,4 @@
+<% @ingredients.each do |ingredient| %>
+<p><%= ingredient.name %>: <%= ingredient.cost %></p>
+<% end %>
+<%= link_to 'Sort Ingredients Alphabetically', "/ingredients?sort=yeahbaby" %>
```
cjsim89, line 4: This does accomplish what the _spirit_ of the user story asks for, but it does just ask for the list to be ordered alphabetically, so we don't need a link that does it. You can alter the `@ingredients` on line 1 to always be ordered after implementing this user story by calling it from the controller, which wouldn't need this query param.

config/routes.rb
```diff
@@ -3,4 +3,6 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "/ingredients", to: 'ingredients#index'
+  get "/recipes/:id", to: 'recipes#show'
```
cjsim89, line 7: Good restful routes! 👍

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      expect(fatback.name).to appear_before(pintos.name)
+      expect(pintos.name).to appear_before(tomatoes.name)
+
+      expect(page).to have_link('Sort Ingredients Alphabetically')
+      click_on 'Sort Ingredients Alphabetically'
+
+      expect(@brisket.name).to appear_before(fatback.name)
+      expect(pintos.name).to appear_before(@potatoes.name)
+      expect(@potatoes.name).to appear_before(tomatoes.name)
+      expect(tomatoes.name).to_not appear_before(@potatoes.name)
```
cjsim89, line 33: This is a great test overall! Good use of objects (lots of 'em), and orderly to help it all along. 👍

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    RecipeIngredient.create!(ingredient: @garlic, recipe: @spaghetti)
+    RecipeIngredient.create!(ingredient: @onion, recipe: @spaghetti)
+    RecipeIngredient.create!(ingredient: @salt, recipe: @spaghetti)
+
+    RecipeIngredient.create!(ingredient: @cornmeal, recipe: @cornbread)
+    RecipeIngredient.create!(ingredient: @butter, recipe: @cornbread)
+    RecipeIngredient.create!(ingredient: @salt, recipe: @cornbread)
+
+    RecipeIngredient.create!(ingredient: @beets, recipe: @borscht)
+    RecipeIngredient.create!(ingredient: @salt, recipe: @borscht)
```
cjsim89, line 40: This was a lot of objects! Not a problem, but if you feel this took you a while to create, you can probably take this data in half for the other ICs this inning and will be fine.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+  describe '/recipes/:id' do 
+    it 'displays name, complexity, genre, ingredients' do 
+      visit "/recipes/#{@spaghetti.id}"
+      expect(page).to have_content(@spaghetti.name)
+      expect(page).to have_content("Complexity: #{@spaghetti.complexity}")
+      expect(page).to have_content("Genre: #{@spaghetti.genre}")
+      expect(page).to have_content("Ingredients:")
+      expect(page).to have_content(@tomatoes.name)
```
cjsim89, line 50: Right here would be a perfect use case for a `within` block - for me, almost anywhere I use a `.each` in a view is where I would put a `within` in my tests, because all the data inside that .each would be dynamic/changing for each page.

spec/models/ingredient_spec.rb
```diff
@@ -12,4 +12,15 @@
...
 
+   describe "alphabetical sort" do 
+    it 'sort alphabetically' do 
+      salt = Ingredient.create!(name: 'salt', cost: 2)
+      onion = Ingredient.create!(name: 'onion', cost: 4)
+      cornmeal = Ingredient.create!(name: 'cornmeal', cost: 10)
+      fatback = Ingredient.create!(name: 'fatback', cost: 8)
+
+      expect(Ingredient.sort_alphabetically('yeahbaby')).to eq([cornmeal, fatback, onion, salt])
+      expect(Ingredient.sort_alphabetically('nobaby')).to eq([salt, onion, cornmeal, fatback])
```
cjsim89, line 23: Really good to see you testing both scenarios here. Keep this up!

spec/models/recipe_spec.rb
```diff
@@ -12,4 +24,10 @@
     it {should have_many :recipe_ingredients}
     it {should have_many(:ingredients).through(:recipe_ingredients)}
   end
+
+  describe "methods" do 
+    it 'total_cost' do 
+      expect(@spaghetti.total_cost).to eq(19)
```
cjsim89, line 30: Nice job with this overall. Might be nice to see a 2nd recipe being tested; maybe one with no ingredients at all, whose `total_cost` would return `0`.

PR #25, opened 2023-05-23
Andy Stilmock IC #1 

AStilmock:
Reflection
This first independent challenge seemed to be an intro to application building where we weren't required to complete all steps of setup and simply de-bug and add a few features to the application. It was a bit tricky but ultimately was a good introduction to how ICs will go in the future for the rest of this module. 

I was able to complete the de-bugging of the associations, completed all three user stories as well as one of the extension stories.

### DIFF
--- a/app/controllers/ingredients_controller.rb
+++ b/app/controllers/ingredients_controller.rb
@@ -1,0 +1,5 @@
+class IngredientsController < ApplicationController
+  def index
+    @ingredients = Ingredient.all
+  end
+end
--- a/app/controllers/ingredients_controller.rb
+++ b/app/controllers/ingredients_controller.rb
@@ -1,5 +1,5 @@
 class IngredientsController < ApplicationController
   def index
-    @ingredients = Ingredient.all
+    @ingredients = Ingredient.all.sort_alpha
   end
 end
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,0 +1,5 @@
+class RecipesController < ApplicationController
+  def show
+    @recipes = Recipe.find(params[:id])
+  end
+end
--- a/app/models/ingredient.rb
+++ b/app/models/ingredient.rb
@@ -1,3 +1,5 @@
 class Ingredient < ApplicationRecord
    validates_presence_of :name, :cost
+   has_many :recipe_ingredients
+   has_many :recipes, through: :recipe_ingredients
 end
--- a/app/models/ingredient.rb
+++ b/app/models/ingredient.rb
@@ -2,4 +2,12 @@ class Ingredient < ApplicationRecord
    validates_presence_of :name, :cost
    has_many :recipe_ingredients
    has_many :recipes, through: :recipe_ingredients
+
+   def self.total_cost
+      sum(:cost)
+   end
+
+   def self.sort_alpha
+      order(name: :asc)
+   end
 end
--- a/app/models/ingredient.rb
+++ b/app/models/ingredient.rb
@@ -7,6 +7,7 @@ class Ingredient < ApplicationRecord
       sum(:cost)
    end
 
+
    def self.sort_alpha
       order(name: :asc)
    end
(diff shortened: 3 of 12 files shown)


app/controllers/ingredients_controller.rb
```diff
@@ -0,0 +1,5 @@
+class IngredientsController < ApplicationController
+  def index
+    @ingredients = Ingredient.all.sort_alpha
```
cjsim89, line 3: This is good! One tidbit; when you do `Ingredient.` anything, it will assume you are doing `.all` in between. So, this could just read `Ingredient.sort_alpha` and it would be doing it on "all". Yours isn't wrong, just want to show you this alternative.

app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,5 @@
+class RecipesController < ApplicationController
+  def show
+    @recipes = Recipe.find(params[:id])
```
cjsim89, line 3: This is fine, your variable probably should be singular since this is referencing one specific Recipe instance.

app/models/ingredient.rb
```diff
@@ -1,3 +1,14 @@
 class Ingredient < ApplicationRecord
    validates_presence_of :name, :cost
+   has_many :recipe_ingredients
+   has_many :recipes, through: :recipe_ingredients
+
+   def self.total_cost
+      sum(:cost)
```
cjsim89, line 7: Be careful with this implementation - the user story says we need to know the total cost of all ingredients _on a recipe_, so if thi sis being used on an association on Recipe, then it would be ok. But if this lets us know total cost of any collection of Ingredients, we need to be sure it's tested in both ways. 

An alternative would be to implement this as an instance method on the Recipe model: 
```ruby
def total_cost
   ingredients.sum(:cost)
end
```

config/routes.rb
```diff
@@ -3,4 +3,8 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/ingredients", to: "ingredients#index"
+  
+  get "/recipes/:id", to: "recipes#show"
```

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      expect(page).to have_content("Cost: #{@beef.cost}")
+
+      expect(page).to have_content("Name: #{@pasta.name}")
+      expect(page).to have_content("Cost: #{@pasta.cost}")
+    end
+
+    it "lists ingredients alphabetically" do
+      visit "/ingredients"
+
+      within "/ingredients"
```
cjsim89, line 31: It's good to use a within block here, but a within takes an HTML identifier (either `#` for id, or `.` for class) and what that identifier is called. You've provided a path, which was probably a typo - try to change this to the id that you created in your view.

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      expect(@beef.name).to appear_before(@oil.name)
+      expect(@oil.name).to appear_before(@pasta.name)
+      expect(@pasta.name).to appear_before(@tomatoes.name)
```
cjsim89, line 34: Good job using orderly to make this test!

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    it "shows recipe attributes" do
+      visit "/recipes/#{@recipe1.id}"
+
+      expect(page).to have_content("Name: #{@recipe1.name}")
+      expect(page).to have_content("Complexity: #{@recipe1.complexity}")
+      expect(page).to have_content("Genre: #{@recipe1.genre}")
+
+      expect(page).to_not have_content("Name: #{@recipe2.name}")
```
cjsim89, line 35: Good use of `.to_not` to differentiate the cases of your test suite - keep this up!

app/models/ingredient.rb
cjsim89: Either way, since you made new model methods you must also test them in `spec/models`. Models should consistently stay at 100% coverage.

app/models/ingredient.rb
```diff
@@ -1,3 +1,14 @@
...
+   end
+
+
+   def self.sort_alpha
+      order(name: :asc)
```
cjsim89, line 12: This is also missing a model test.

PR #26, opened 2023-05-23
Alex Kiselich 

AlexKiselich:
I think I did ok, I stumbled in the very first part of the relationships but once I took a deep breath and calmed down a little, I re-focused and got into a good grove.

### DIFF
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -4,4 +4,10 @@ class RecipesController < ApplicationController
     @recipe_ingredients = @recipe.ingredients
     @total_cost = @recipe.total_cost_ingredients
   end
+
+  def update
+    @recipe = Recipe.find(params[:id])
+
+    
+  end
 end
--- a/app/controllers/recipes_controller.rb
+++ b/app/controllers/recipes_controller.rb
@@ -1,0 +1,7 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @recipe_ingredients = @recipe.ingredients
+    @total_cost = @recipe.total_cost_ingredients
+  end
+end
--- a/app/models/ingredient.rb
+++ b/app/models/ingredient.rb
@@ -1,3 +1,6 @@
 class Ingredient < ApplicationRecord
    validates_presence_of :name, :cost
+
+   has_many :recipe_ingredients
+   has_many :recipes, through: :recipe_ingredients
 end
--- a/app/models/ingredient.rb
+++ b/app/models/ingredient.rb
@@ -3,4 +3,8 @@ class Ingredient < ApplicationRecord
 
    has_many :recipe_ingredients
    has_many :recipes, through: :recipe_ingredients
+
+   def self.alphabetically
+      order(name: :asc)
+   end
 end
--- a/app/models/recipe.rb
+++ b/app/models/recipe.rb
@@ -3,4 +3,8 @@ class Recipe < ApplicationRecord
    
    has_many :recipe_ingredients
    has_many :ingredients, through: :recipe_ingredients
+
+   def total_cost_ingredients
+      ingredients.sum(:cost)
+   end
 end
--- a/app/models/recipe.rb
+++ b/app/models/recipe.rb
@@ -1,3 +1,6 @@
 class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
+   
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
 end
--- a/app/views/recipes/show.html.erb
+++ b/app/views/recipes/show.html.erb
@@ -7,3 +7,9 @@
   <p><%= ingred.name %></p>
 <% end %>
 <h3>Total Cost of Ingredients: <%= @total_cost %></h3>
+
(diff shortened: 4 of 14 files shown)


app/controllers/recipes_controller.rb
```diff
@@ -0,0 +1,13 @@
+class RecipesController < ApplicationController
+  def show
+    @recipe = Recipe.find(params[:id])
+    @recipe_ingredients = @recipe.ingredients
+    @total_cost = @recipe.total_cost_ingredients
```
cjsim89, line 5: These 2 instance variables are not really necessary; you can call `.ingredients` and `total_cost_ingredients` from the view itself, which would help keep this controller method "skinny".

app/models/ingredient.rb
```diff
@@ -1,3 +1,10 @@
 class Ingredient < ApplicationRecord
    validates_presence_of :name, :cost
+
+   has_many :recipe_ingredients
+   has_many :recipes, through: :recipe_ingredients
+
+   def self.alphabetically
+      order(name: :asc)
```
cjsim89, line 8: Good AR, and good use of a class method!

app/models/recipe.rb
```diff
@@ -1,3 +1,10 @@
 class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
+   
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
+
+   def total_cost_ingredients
+      ingredients.sum(:cost)
```
cjsim89, line 8: Great job with this AR!

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,15 @@
...
+<p>Complexity: <%= @recipe.complexity %></p>
+<p>Genre: <%= @recipe.genre %></p>
+
+<h2>Ingredients List</h2>
+<% @recipe_ingredients.each do |ingred| %>
+  <p><%= ingred.name %></p>
+<% end %>
+<h3>Total Cost of Ingredients: <%= @total_cost %></h3>
+
+<%= form_with url: "/recipes/#{@recipe.id}", method: :patch, local: true do |f| %>
```
cjsim89, line 11: Looks like you were working on the 2nd extension. This is a good start, but I'm not sure we're really "updating a specific recipe" - the user story asks us to make a new relationship between an existing recipe (@recipe.id) and and the _id of an ingredient_ that the user enters in the form. So, we wouldn't need a `:name` field in the form, rather an `:ingredient_id` field. 

Also, in terms of a restful route, our URL should reflect that wer'e adding an ingredient to a recipe; so one of these options: 
```
"/recipes/:id/ingredients"
"/recipe_ingredients"
```
The 2nd option would mean we'd need a hidden field on the form to send in the recipe_id as well as the ingredient_id, but eitehr way would work. 

Try finishing this out for good practice!

config/routes.rb
```diff
@@ -3,4 +3,9 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/ingredients", to: "ingredients#index"
+
+  get "/recipes/:id", to: "recipes#show"
+  patch "/recipes/:id", to: "recipes#update"
```
cjsim89, line 10: Following the advice above, this route woudl change slightly; patching is for updating, so it's good your verb & action match, but consider again what the story is asking us to do: create a relationship.... What would change here?

spec/features/ingredients/index_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    it 'Displays a list of all the ingrediants including their name and cost' do
+      visit "/ingredients"
+      expect(page).to have_content(@milk.name)
+      expect(page).to have_content(@milk.cost)
+      expect(page).to have_content(@flour.name)
+      expect(page).to have_content(@flour.cost)
+    end
+    it 'displays the list of ingredients alphabetically' do
+      visit "/ingredients"
+      expect(@flour.name).to appear_before(@milk.name)
```
cjsim89, line 18: Good use of orderly in this test!

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      visit "/recipes/#{@cake.id}"
+      expect(page).to have_content("Total Cost of Ingredients: 3")
+    end
+
+    it 'displays a form to add a new ingredient to this recipe' do
+      visit "/recipes/#{@cake.id}"
+      save_and_open_page
+      expect(page).to have_field("Add Ingredient")
+
+      fill_in("Add Ingredient", with: "Sugar")
```
cjsim89, line 41: Read the user story carefully, it asks us for the ID of an ingredient, not a name.

spec/models/recipe_spec.rb
```diff
@@ -1,15 +1,28 @@
...
-
   describe "relationships" do
     it {should have_many :recipe_ingredients}
     it {should have_many(:ingredients).through(:recipe_ingredients)}
   end
+  
+  describe "#instance_Methods" do
+    describe '#total_cost_ingredients' do
+      it 'shows the total cost of the ingrediants in the recipe' do
+        expect(@cake.total_cost_ingredients).to eq(3)
```
cjsim89, line 24: For these types of model tests it's a good idea to have a 2nd object to test off of - a 2nd recipe for example that has no ingredients, whose `total_cost_ingredients` method would return `0`.

PR #27, opened 2023-05-24
Matthew Johnson

MWMJohnson:
Happy I got through user story 3, but wish I at least completed one extension.

### DIFF
--- a/app/models/recipe.rb
+++ b/app/models/recipe.rb
@@ -1,3 +1,6 @@
 class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
+
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
 end
--- a/app/models/recipe.rb
+++ b/app/models/recipe.rb
@@ -3,4 +3,9 @@ class Recipe < ApplicationRecord
 
    has_many :recipe_ingredients
    has_many :ingredients, through: :recipe_ingredients
+
+   def total_cost
+      ingredients.sum(:cost)
+   end
+
 end
--- a/app/views/recipes/show.html.erb
+++ b/app/views/recipes/show.html.erb
@@ -9,4 +9,4 @@
     <% end %>
     </ul>
 </ul>
-<h2>Total Cost: $ <%= @ingredients_cost %></h2>
+<h2>Total Ingredients Cost: $ <%= @ingredients_cost %></h2>
--- a/app/views/recipes/show.html.erb
+++ b/app/views/recipes/show.html.erb
@@ -1,0 +1,12 @@
+<h1>Recipe Name: <%= @recipe.name %></h1>
+<ul>
+  <li>Complexity: <%= @recipe.complexity %></li>
+  <li>Genre: <%= @recipe.genre %></li>
+  <li>Ingredients:</li>
+    <ul>
+    <% @recipe.ingredients.each do |ingredient| %>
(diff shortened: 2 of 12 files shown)


app/models/recipe.rb
```diff
@@ -1,3 +1,11 @@
 class Recipe < ApplicationRecord
    validates_presence_of :name, :complexity, :genre
+
+   has_many :recipe_ingredients
+   has_many :ingredients, through: :recipe_ingredients
+
+   def total_cost
+      ingredients.sum(:cost)
```
cjsim89, line 8: Good use of AR here!

config/routes.rb
```diff
@@ -3,4 +3,8 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/ingredients", to: "ingredients#index"
+  
+  get "/recipes/:id", to: "recipes#show"
```
cjsim89, line 9: Nice restful routes!

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,12 @@
...
+  <li>Complexity: <%= @recipe.complexity %></li>
+  <li>Genre: <%= @recipe.genre %></li>
+  <li>Ingredients:</li>
+    <ul>
+    <% @recipe.ingredients.each do |ingredient| %>
+      <li> <%= ingredient.name %></li>
+    <% end %>
+    </ul>
+</ul>
+<h2>Total Ingredients Cost: $ <%= @ingredients_cost %></h2>
```
cjsim89, line 12: This is fine, but you can also call `.total_cost` directly here instead of making another instance variable, which would help keep your controllers "skinny".

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  
+    visit "/recipes/#{@recipe1.id}"
+  end
+
+  it "lists all the attributes of the recipe along with their ingredients" do
+    expect(page).to have_content("Name: #{@recipe1.name}")
+    expect(page).to have_content("Complexity: #{@recipe1.complexity}")
+    expect(page).to have_content("Genre: #{@recipe1.genre}")
+
+    expect(page).to have_content(@beef.name)
```
cjsim89, line 27: This would be a good opportunity to use a `within` block - anytime I use a `.each` in a view, I think about using a within, because that .each is going to be iterating over dynamic (always different) data.

spec/models/recipe_spec.rb
```diff
@@ -12,4 +11,30 @@
...
+        @recipe1_ingredient3 = RecipeIngredient.create!(recipe_id: @recipe1.id, ingredient_id: @flour.id)
+    
+        @recipe2_ingredient1 = RecipeIngredient.create!(recipe_id: @recipe2.id, ingredient_id: @beef.id)
+        @recipe2_ingredient2 = RecipeIngredient.create!(recipe_id: @recipe2.id, ingredient_id: @flour.id)
+      
+      end
+      it "finds the total cost of ingredients for the recipe" do
+        expect(@recipe1.total_cost).to eq(12)
+
+        expect(@recipe2.total_cost).to eq(8)
```
cjsim89, line 36: Good job with this model test! Also, all of your object/data setup for each test, both model and feature, are great! Keep this up 🙌

cjsim89 (commented):
Really solid job Matthew! Keep up the good work. I also recommend trying out the extensions for more practice. 👍

PR #30, opened 2023-05-26
Myles Nottingham

MylesNottingham:
This felt like it went well. Sorry so late 😬


app/controllers/ingredients_controller.rb
```diff
@@ -0,0 +1,5 @@
+class IngredientsController < ApplicationController
+  def index
+    @ingredients = Ingredient.order_by_name
```
epintozzi, line 3: Nice job keeping the logic out of the controller!

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+
+    it "lists the recipe's ingredients" do
+      expect(page).to have_content("Ingredients:")
+
+      within "#ingredients" do
+        expect(page).to have_content(ingredient_1.name)
+        expect(page).to have_content(ingredient_2.name)
+        expect(page).to have_content(ingredient_3.name)
+        expect(page).to have_content(ingredient_4.name)
```
epintozzi, line 39: I might have created one more ingredient that does NOT belong to this recipe to ensure ONLY the ingredients belonging to the recipe show up and that this isn't a false positive (I don't think it is - but this would give us more confidence in that).

epintozzi: Nevermind. I see you do exactly this in a later spec. Nice work!

app/views/recipes/show.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    <% end %>
+  </ul>
+<% end %><br/>
+
+<h3>Total Cost: $<%= @recipe.total_cost %></h3><br/>
+
+<form action="/recipes/<%= @recipe.id %>" method="post">
+  <label for="ingredient_id">Add an Ingredient to this Recipe by Ingredient ID</label>
+  <input type="number" name="ingredient_id" id="ingredient_id">
+  <input type="submit" value="Submit">
```
epintozzi, line 19: Excellent work getting through the extensions. I would like to see you start getting more comfortable with Rails built in form helpers, but this absolutely works.

spec/features/recipes/show_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+      expect(page).to have_field(:ingredient_id)
+      expect(page).to have_button("Submit")
+
+      fill_in :ingredient_id, with: ingredient_1.id
+      click_button "Submit"
+
+      expect(current_path).to eq("/recipes/#{recipe_2.id}")
+
+      expect(page).to have_content(ingredient_1.name)
+      expect(page).to have_content("Total Cost: $6")
```
epintozzi, line 86: We would expect the calculation to update when we add an ingredient (assuming we coded it correctly), but I like that you're verifying we see that change here as well. There are ways that code could be altered such that it wouldn't be a guarantee, and this test would catch that if such a change were ever made.

epintozzi (commented):
Nice work Myles!

